PR #145, opened 2026-03-27
GraviScan 7.1/7: Parallelize scanner subprocess initialization

### LINKED ISSUE #144: GraviScan: Sequential scanner init causes ~5min startup with 4 scanners
## Problem

`ScanCoordinator.initialize()` spawns scanner subprocesses **sequentially** — each Python worker must fully complete `sane.init()` + `sane.open()` + geometry setup before the next one starts. With 4 scanners this takes ~5 minutes at startup.

This was originally done to prevent SANE init contention, since `sane_init()` is a process-global C call. However, each scanner already runs in its **own subprocess** (separate Python process), so the global state is fully isolated — the sequential constraint is unnecessary.

## Expected behavior

All 4 subprocesses should spawn in parallel via `Promise.allSettled`. Total init time should be ~1 minutes (the time of the slowest single scanner), not the sum of all.

## Additional context

- The 5-second USB stagger during `scanOnce()` is a separate, valid constraint (epkowa `device.start()` contention) and should NOT change
- If one scanner fails (e.g. unplugged), the others should still initialize successfully
- After test scan, subprocesses stay alive and are reused for the real scan — this reuse path should be preserved

**File:** `src/main/scan-coordinator.ts` — the `initialize()` method

### LINKED ISSUE #185: GraviScan: Parallel subprocess initialization
## Context

In `scan-coordinator.ts`, scanner subprocesses are initialized sequentially (one at a time). Issue #144 argues this is unnecessary since each subprocess runs in its own Python process with isolated SANE context — there is no process-wide global state to contend with.

## Problem

Sequential initialization causes a startup penalty proportional to the number of scanners. With 4 scanners, initialization takes ~20 seconds (5s per scanner) instead of ~5 seconds in parallel.

## Proposal

Switch `ScanCoordinator.initialize()` to spawn subprocesses in parallel using `Promise.allSettled()`. This requires designing error semantics for partial init failure:

- What happens if 3/4 scanners initialize successfully but one fails?
- Should the coordinator proceed with the successfully initialized scanners?
- Should it retry the failed scanner?
- How is the partial failure communicated to the UI?

## References

- #144 (original issue reporting sequential init as unnecessary)
- `src/main/graviscan/scan-coordinator.ts` — `initialize()` method
- Design note in `openspec/changes/add-graviscan-coordinator-subprocess/design.md` section 4

blm3886:
## Summary

- Refactors `ScanCoordinator.initialize()` to spawn all scanner subprocesses in parallel via `Promise.allSettled` instead of sequentially awaiting each one
- Failed scanners are removed from the map and the rest proceed — one unplugged scanner no longer blocks startup
- Emits `scanner-init-status` events per scanner so the UI can show individual connection progress
- Adds 8 unit tests covering parallel spawn, partial failure, reuse, stale cleanup, and all-fail scenarios

Closes #144
Closes #185

## Why

Each subprocess is an independent OS process with its own SANE context targeting a distinct USB device. The sequential init was an overly cautious measure against SANE global state contention, but since each scanner runs in a **separate Python process**, the global state is fully isolated. With 4 scanners, sequential init takes ~5 min; parallel should be ~1-2 min.

Only review commits after 19020e9:

- `0763722` perf: parallelize scanner subprocess initialization

Diff for this PR only: [graviscan/7-renderer-ui...perf/parallelize-scanner-init](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/compare/graviscan/7-renderer-ui...perf/parallelize-scanner-init)

## Test plan

- [x] Unit tests: 8/8 passing (parallel timing, partial failure, reuse, stale cleanup, all-fail)
- [x] Hardware validation: 4 Epson scanners on Linux USB — pending
- [x] CI test pass 

### DIFF
--- a/src/main/scan-coordinator.ts
+++ b/src/main/scan-coordinator.ts
@@ -2,7 +2,7 @@
  * Scan Coordinator
  *
  * Orchestrates multiple ScannerSubprocess instances for parallel scanning.
- * Handles staggered subprocess startup, simultaneous scan triggers,
+ * Handles parallel subprocess initialization, simultaneous scan triggers,
  * interval/continuous mode timing, and cleanup.
  *
  * Usage:
@@ -78,9 +78,10 @@ export class ScanCoordinator extends EventEmitter {
   }
 
   /**
-   * Staggered initialization: spawn subprocesses one at a time,
-   * waiting for each to signal ready before starting the next.
-   * This prevents SANE init contention.
+   * Parallel initialization: spawn all subprocesses concurrently.
+   * Each subprocess is an independent OS process with its own SANE context
+   * targeting a distinct physical USB device, so no sequential ordering is needed.
+   * Reuse/cleanup of existing subprocesses is done first (sequential, fast).
    */
   async initialize(scanners: ScannerConfig[]): Promise<void> {
     this.state = 'initializing';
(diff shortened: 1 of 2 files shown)


eberrigan:
Superseded by #357, which shipped the same parallelized scanner subprocess initialization plus a fix for the re-entrancy race that this PR didn't address. Closing as superseded.

Issue #130, opened 2026-03-19
GraviScan 4/7: Main process — IPC handlers, coordinator, subprocess


eberrigan:
**Parent epic:** #126
**Depends on:** #129 (types + Python)
**Plan section:** PR 4 in `docs/superpowers/plans/2026-03-18-graviscan-merge-strategy.md`

## Scope

All main-process code: IPC handler registration, multi-scanner coordination, subprocess management, USB detection, logging, and the preload context bridge.

### Files
- `src/main/graviscan-handlers.ts` (1,225 lines) — 10 IPC handlers
- `src/main/scan-coordinator.ts` (442 lines) — multi-scanner orchestration with staggered init
- `src/main/scanner-subprocess.ts` (301 lines) — per-scanner Python process management
- `src/main/lsusb-detection.ts` (209 lines) — USB scanner enumeration
- `src/main/scan-logger.ts` (86 lines) — persistent scan event log
- `src/main/graviscan-path-utils.ts` (50 lines) — output directory formatting
- `src/main/main.ts` (+346/-311) — handler registration, process lifecycle
- `src/main/preload.ts` (+270/-11) — GraviScan API context bridge
- `src/main/config-store.ts` (+10/-1)
- `src/main/database.ts` (+115/-9)
- `src/main/database-handlers.ts` (+833/-124) — GraviScan DB operations

### New IPC Handlers (13 total)
`graviscan:detect-scanners`, `graviscan:get-config`, `graviscan:validate-config`, `graviscan:get-scan-status`, `graviscan:mark-job-recorded`, `graviscan:cancel-scan`, `graviscan:get-output-dir`, `graviscan:read-scan-image`, `graviscan:upload-all-scans`, `graviscan:platform-info`, `scanner:scan`, `scanner:get-scanner-id`, `session:get/reset`

### Critical Review Points
- [ ] `main.ts` restructuring preserves all existing handlers (camera, DAQ, scanner, config, session, db)
- [ ] `database-handlers.ts` additions follow existing pattern (try/catch, DatabaseResponse<T>)
- [ ] No `any` types in new handler signatures
- [ ] Process lifecycle: `before-quit` cleans up GraviScan subprocesses
- [ ] Preload bridge: no overexposure of Node APIs

### Acceptance Criteria
- [ ] `npx tsc --noEmit` passes
- [ ] `npm run test:unit` passes (no regressions to existing tests)
- [ ] `npm run lint` passes
[...]

Issue #126, opened 2026-03-19
Epic: Merge GraviScan (#123) — Incremental Integration


eberrigan:
## Overview

PR #123 (`feature/graviscan`) adds a complete multi-scanner plant phenotyping system (~20K LOC, 105 files). It is too large to review and merge as a single PR. This epic tracks splitting it into 7 incremental PRs that can each pass CI independently and be reviewed in isolation.

**Original PR:** #123
**Author:** @blm3886
**Base branch:** `dev`
**Plan document:** `docs/superpowers/plans/2026-03-18-graviscan-merge-strategy.md`

## Platform Strategy

| Platform | Scanner Backend | Status | Focus |
|----------|----------------|--------|-------|
| **Linux** | SANE (python-sane) | Fully implemented | **Primary — lab machines** |
| **macOS** | Mock mode only | Dev/test only | Development + CI |
| **Windows** | TWAIN (pytwain) | Not yet implemented | **Future phase** |

## Merge Order (dependency chain)

Each PR depends on the one above it. Merge sequentially.

- [ ] #127 — **PR 1:** Database schema & migrations (14 migrations, 8 new models)
- [ ] #128 — **PR 2:** CI & build infrastructure (APP_MODE, NPM_TOKEN, icons, forge config)
- [ ] #129 — **PR 3:** Types + Python SANE backend + tests (types, scan_worker, scan_regions)
- [ ] #130 — **PR 4:** Main process — IPC handlers, coordinator, subprocess management
- [ ] #131 — **PR 5:** Main process — Box backup (rclone) & Supabase upload
- [ ] #132 — **PR 6:** Renderer — hooks & state logic + unit tests
- [ ] #133 — **PR 7:** Renderer — components, pages, routes + E2E tests

## Known Issues to Address During Split

1. `python/tests/test_scans.py` is empty (vestigial) — real tests are in `test_scan_worker.py` and `test_scan_regions.py`
2. `GraviScan.tsx` is 3,464 lines — must be refactored into smaller components
3. No unit tests for renderer hooks (2,494 LOC, 0 tests)
4. No automated E2E tests (interactive test only)
5. Windows TWAIN not implemented (future phase)
6. `CLAUDE.md` changes in PR should NOT be merged — managed by OpenSpec

## Testing Strategy

| Layer | Framework | Threshold | CI Platform |
|-------|-----------|-----------|-------------|
[...]

PR #186, opened 2026-04-10
feat: GraviScan coordinator + subprocess — Increment 3b

eberrigan:
## Summary

- Cherry-pick and adapt `scan-coordinator.ts`, `scanner-subprocess.ts`, and `scan-logger.ts` from PR #138 into `src/main/graviscan/`
- Move `PlateConfig` and `ScannerConfig` types from session-handlers.ts local definitions into `src/types/graviscan.ts` (shared IPC-facing types)
- `ScanCoordinator` explicitly `implements ScanCoordinatorLike` for compile-time contract enforcement
- 47 new unit tests across 3 test files (scan-coordinator, scanner-subprocess, scan-logger)
- No IPC wiring — modules are isolated and tested, ready for Increment 3c

### Adaptations from Ben's branch (not cherry-picked as-is)

- Rename failures surface as `rename-error` events instead of silent `console.error`
- File existence + non-zero size verification after `scan-complete`
- USB stagger delay logged via `scanLog()` with scanner ID and delay duration
- Log retention configurable via `GRAVISCAN_LOG_RETENTION_DAYS` env var (default 180 days)
- Sleep/cancel race condition fixed — `cancelAll()` resolves pending sleep promise
- Dead `CoordinatorEvent` type removed
- `USB_STAGGER_DELAY_MS` extracted as named module-level constant

### Related issues

- Part of GraviScan integration epic #126
- Source: PR #138 (branch `origin/graviscan/4-main-process`), issue #130
- Filed #185 for deferred parallel subprocess initialization (#144)

### OpenSpec proposal

`openspec/changes/add-graviscan-coordinator-subprocess/` — reviewed by 5 specialized subagents

## Test plan

- [x] `npx tsc --noEmit` — zero type errors
- [x] `npx eslint` on all changed files — zero errors
- [x] `npx prettier --check` — all files formatted
- [x] `npm run test:unit` — 631 passed, 9 skipped (0 regressions)
- [ ] CI: lint, compilation, unit tests, Python tests
- [ ] E2E: N/A (no IPC wiring or UI changes in this increment)
- [ ] Packaging: N/A (modules not yet reachable from app)

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -56,6 +56,33 @@ This document describes all configurable constants and default values in the Blo
 - **Gain**: Use only when lighting is insufficient (adds noise)
 - **Gamma**: Adjust for non-linear brightness correction (typically 1.0-2.2)
 
+## GraviScan Settings
+
+### Scan Coordination
+
+| Constant               | Type     | Default | Description                                                                                             | Location                                                                            | Admin Configurable          |
+| ---------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------- | --------------------------- |
+| `USB_STAGGER_DELAY_MS` | `number` | `5000`  | Delay (ms) between starting scanners within a grid row. Prevents epkowa SANE backend USB contention.    | [src/main/graviscan/scan-coordinator.ts](../src/main/graviscan/scan-coordinator.ts) | ❌ No - Hardware constraint |
+| `SCAN_ROW_TIMEOUT_MS`  | `number` | `90000` | Per-row scan timeout (ms). If a subprocess doesn't respond within this window, it is treated as failed. | [src/main/graviscan/scan-coordinator.ts](../src/main/graviscan/scan-coordinator.ts) | ❌ No - Safety timeout      |
+
+### Logging
+
+| Constant             | Type     | Default | Description                                                                                                                      | Location                                                                  | Admin Configurable   |
+| -------------------- | -------- | ------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------- | -------------------- |
+| `LOG_RETENTION_DAYS` | `number` | `180`   | Days to retain scan log files before cleanup. Default 180 days for scientific workflows where analysis/publication takes months. | [src/main/graviscan/scan-logger.ts](../src/main/graviscan/scan-logger.ts) | ✅ Yes - via env var |
+
+### Environment Variables
+
+| Variable                       | Default | Description                                                                                                               |
+| ------------------------------ | ------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `GRAVISCAN_LOG_RETENTION_DAYS` | `180`   | Override log retention period (days). Set before app launch. Must be a positive integer; invalid values fall back to 180. |
+
+**Example:**
+
+```bash
+GRAVISCAN_LOG_RETENTION_DAYS=365 npm start
+```
+
 ## File Locations
 
 All default settings are defined in TypeScript type definition files:
--- a/openspec/changes/add-graviscan-coordinator-subprocess/design.md
+++ b/openspec/changes/add-graviscan-coordinator-subprocess/design.md
@@ -1,0 +1,87 @@
+# Design: GraviScan Coordinator + Subprocess
+
+## Overview
+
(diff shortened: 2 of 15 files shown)


eberrigan (commented):
> **Verdict: REQUEST_CHANGES** (posted as comment — GitHub does not allow requesting changes on your own PR)

## Review Summary

Solid incremental work — the module structure, type reconciliation, and Copilot feedback loop are well-executed. However, the behavioral reviewers found several state machine bugs (cancel/shutdown race conditions, stuck 'waiting' state) and the testing reviewer flagged real timer usage that adds ~30s of dead time to CI. These should be fixed before merge.

## Blocking Issues

### B1. `scanInterval` leaves state as `'waiting'` after `cancelAll()` during sleep

When `cancelAll()` fires during the inter-cycle sleep, the sleep resolves, the while-loop breaks, `scanInterval` emits `interval-complete` and returns — but **nobody sets state back to `'idle'`**. `isScanning` returns `true` forever.

**Fix:** Set `this.state = 'idle'` in `scanInterval()` before emitting `interval-complete` (or after the while loop exits).

### B2. No timeout on `Promise.all` in `scanOnce` — hang risk

If a subprocess is stuck (SANE hardware hang, broken pipe), the per-row `Promise.all` at line ~297 waits forever. `cancelAll()` sends cancel to stdin but if the process doesn't respond, there's no escape.

**Fix:** Add a per-row timeout that rejects remaining promises if subprocesses don't respond within a reasonable window (e.g., `SCAN_TIMEOUT_MS`). Or listen for the coordinator's `cancelled` flag to reject pending promises.

### B3. `scan_ended_at: null` injected into forwarded scan events

The `sub.on('event', ...)` listener (line ~139) captures `this.currentGridEndedAt` at event-fire time. For `scan-complete` events, this is still `null` because the row hasn't finished yet. Every forwarded scan-complete event carries `scan_ended_at: null`. This is misleading metadata.

**Fix:** Either (a) stop injecting `scan_ended_at` into per-plate events (only include it in `grid-complete`), or (b) emit enriched events after the row completes.

### B4. Real 5s stagger delays in unit tests — 30s CI waste

Three tests (`stagger`, `partial failure`, `cancel during interval`) use real `setTimeout` for the USB stagger delay, adding ~30s to every CI run. The `scanInterval` test in the same file already demonstrates `vi.useFakeTimers()`.

**Fix:** Use fake timers or mock `USB_STAGGER_DELAY_MS` to 0 in multi-scanner tests.

### B5. Flaky cancel-during-interval test

```typescript
await new Promise((r) => setTimeout(r, 200));
coordinator.cancelAll();
```
This hopes 200ms is enough for the first cycle to complete. On loaded CI, it may not be.

**Fix:** Use fake timers with deterministic advancement.

## Important Issues

### I1. Regex-based path rewriting (`_st_`, `_cy`) matches first occurrence — could hit directory names

The `replace(/(\d{8}T\d{6})/, stTimestamp)` replaces the first match in the entire path. If the parent directory contains a similar pattern, the wrong segment is rewritten. **Fix:** Apply regex to `path.basename` only, then rejoin.

### I2. `scanOnce` doesn't check `this.cancelled` after `Promise.all` returns

After all subprocesses complete, `scanOnce` proceeds to file verification even if cancel was requested mid-scan. Spurious `grid-complete` events emit with potentially incomplete data after shutdown.

### I3. Synchronous fs calls (`existsSync`, `statSync`, `renameSync`) block Electron main process

Use `fs.promises.*` to avoid UI freezes, especially with multiple scanners/grids.

### I4. `process-error` event silently swallowed by coordinator

Coordinator listens for `event` and `exit` on subprocesses, but not `process-error`. A subprocess spawn failure after initialization is invisible.

### I5. `scan()` sets state to `scanning` even if `sendCommand` fails (proc is null/killed)

State machine inconsistency — subprocess thinks it's scanning but no command was sent.

### I6. `stderr` readline interface never closed — resource leak

Store `stderrRl` as a field and close in `shutdown()`/`kill()`.

### I7. Missing test coverage: `killAll()`, double-spawn guard, zero-size file, 4grid row-grouping, overtime event, cancel during active scanOnce

## Suggestions

- Extract `USB_STAGGER_DELAY_MS` as configurable (env var), like log retention
- Use `Set` for grid index dedup instead of `Array.includes`
- Extract file-verification-and-rename block into a private method
- Add spawn timeout (configurable, e.g., 120s) to prevent infinite hang
- Use filename-embedded date for log retention instead of `mtime`
- Type `ScanWorkerEvent.type` as a string union for compile-time safety
- Document that `initialize()` must not be called concurrently

---

PR #196, opened 2026-04-16
feat: GraviScan renderer (complete) + scanner-config save flow fix

eberrigan:
## Summary

Implements the full GraviScan renderer layer (hooks, pages, routes, E2E) plus three follow-up critical bug fixes surfaced during local smoke testing.

**Three OpenSpec proposals land in this PR:**
1. `add-graviscan-renderer-pages` — Sections 0–15: foundation, data persistence, hooks, UI, routing, E2E (83/83 tasks complete)
2. `fix-scanner-config-save-flow` — Silent-save bug fix + metadata traceability + DB/UI consistency (36/40 tasks complete; 4 pending are deferred manuals or blocked by #206)
3. `surface-disabled-scanners-on-detect` — Catch-22 fix where disabled scanners had no UI path back to enabled (19/20 tasks complete)

All three proposals pass `openspec validate --strict`.

---

## Part 1 — GraviScan renderer (`add-graviscan-renderer-pages`)

- **Foundation** (Sections 0–1): directories, ESLint overrides, test mocks/fixtures, E2E helpers, main-process data persistence layer (`scan-persistence.ts` — crash-safe, follows CylinderScan pattern)
- **DB CRUD handlers**: 7 new IPC handlers for GraviScan reads + plate assignment CRUD
- **Hooks** (Sections 2–8): ported 7 hooks from Ben's pilot branch (`useScannerConfig`, `useScanSession`, `usePlateAssignments`, `useContinuousMode`, `useTestScan`, `useWaveNumber`, metadata validation) and adapted to current IPC/preload API
- **UI pages** (Sections 9–12): ScannerConfig, Metadata, GraviScan, BrowseGraviScans + routes + navigation
- **Integration** (Sections 13–15): bug fixes (#154 file path, #159 start-scan gate), routing tests, path-utils warnings

Fixes #154, #195. Related: #132, #133, #159, #194.

## Part 2 — Scanner Config save flow fix (`fix-scanner-config-save-flow`)

Surfaced during local smoke testing: the Save button silently wrote zero scanners to the DB, leaving downstream Metadata with Prisma FK errors. Adversarial review (5 rounds) plus smoke testing surfaced additional defects (usb_device instability, metadata.json UUID fallback, UI/DB enabled divergence, re-entrancy) — all fixed in this PR.

Root-cause was a placeholder unit test (`expect(fn).toHaveBeenCalled()` with no payload assertion). Lint-rule prevention tracked in #200.

**Key changes:**
[...]

PR #218, opened 2026-05-01
fix: scanner files overwriting each other when system_name is set

blm3886:
## Summary

When multiple scanners on the same machine scanned in parallel, only one image per plate index was saved — the others were silently overwritten because all scanners used the same filename tag.

## Root cause

```typescript
const systemTag = platformInfo?.system_name || \`S\${scannerIdx + 1}\`;
const filename = \`...\${timestamp}_cy1_\${systemTag}_\${plateIndex}.tif\`;
```

If \`GRAVISCAN_SYSTEM_NAME\` env var is set (configured in machine config), \`systemTag\` is the same value for every scanner. All scanners writing the same plate index produce the same filename → overwrites.

User reported "only 3 images per folder" with 3 scanners × 4 plates × multiple cycles — only the last scanner's images per plate survived.

## Fix

Always include a scanner index in the filename, regardless of \`system_name\`. Use \`Sc\` prefix (scanner) to avoid colliding with the metadata \`S\` prefix used for plate sections (S1, S2, S3, S4 — sections within a plate).

New pattern:
- Without \`system_name\`: \`{exp}_st_{ts}_cy{n}_Sc1_00.tif\`
- With \`system_name='GS1'\`: \`{exp}_st_{ts}_cy{n}_GS1_Sc1_00.tif\`

## Files

- \`src/renderer/hooks/useScanSession.ts\`

## Dependencies

- **Depends on Elizabeth's renderer PR #196** — \`useScanSession.ts\` doesn't exist on main yet.

## Will rebase

Once #196 lands, rebase. Diff shrinks to ~5 lines.

## Test plan

- [x] Run continuous scan with 3 scanners on the graviscan machine → all images present, none overwritten
- [x] TypeScript compiles clean


blm3886:
Closing — already has this fix on main (origin/main:src/main/graviscan/scan-coordinator.ts)

PR #255, opened 2026-07-23
fix: add .gitattributes to normalize line endings (CRLF/LF)

### LINKED ISSUE #253: Add .gitattributes to fix line-ending (CRLF/LF) noise on Windows
## Problem

The repo has no `.gitattributes` file, and Windows checkouts commonly have `core.autocrlf=true` set locally. With that combination, `git status` reports large numbers of files as modified purely from stat/line-ending state, even when there is no real content change.

Observed while fixing review findings on #252 (docs-only PR touching 8 files): after running `npm run format`, `git status --short` showed **~300+ files across the entire repo** as modified. `git diff --stat` confirmed only the 8 intentionally-edited files had real content changes — everything else was CRLF/LF noise from `core.autocrlf=true` with no `.gitattributes` to pin line-ending behavior.

This is a footgun: it's easy to accidentally `git add -A` or `git add .` and stage a huge, noisy, unintended diff, especially after running a formatter or IDE action that touches file mtimes.

## Proposed fix

Add a `.gitattributes` file at the repo root:

\`\`\`gitattributes
* text=auto eol=lf
\`\`\`

This normalizes all text files to LF regardless of each contributor's local `core.autocrlf` setting, matching what Prettier writes by default and what the Linux CI runners already use. It only affects future checkouts/commits — it does not itself rewrite any current file content. A follow-up `git add --renormalize .` would be a separate, deliberate step if the team wants to also clean up any files already checked in with inconsistent line endings (that would produce a whitespace-only bulk diff and should be its own PR, reviewed as such).

## Why this matters for this repo specifically

- Windows is one of the three CI platforms and a real dev environment (per `.claude/commands/validate-env.md` / `run-ci-locally.md`), so this isn't a hypothetical edge case.
- A noisy `git status` makes it harder to spot genuinely unintended changes before committing — a real risk in a shared-lab codebase where accidental scope creep in a commit is costly to review.

eberrigan:
## Summary

Add `.gitattributes` to normalize line endings to LF, eliminating the CRLF/LF `git status` noise that Windows checkouts with `core.autocrlf=true` produce.

## Changes

- Add `.gitattributes` at repo root with `* text=auto eol=lf` to pin all text files to LF regardless of a contributor's local `core.autocrlf` setting
- Add explicit `binary` overrides for tracked binary assets that are actually in the repo: `*.png`, `*.db`, `*.xlsx` (test fixtures, sample scan images, SQLite databases, Excel fixtures)
- No other files touched — this does **not** run `git add --renormalize .`; any bulk re-normalization of already-committed file contents is intentionally left as a separate follow-up PR

## OpenSpec Change

No OpenSpec change: small, low-risk infra/config fix (repo-root `.gitattributes` only), no behavior or capability change.

## Testing

Verification performed (see issue #253 for the exact repro this fixes):

- [x] Fresh clone of this branch + `npm install` + `npm run format`: `git status --short` initially shows a batch of files flagged "modified" — this is a pre-existing git-for-windows stat-cache display quirk unrelated to `.gitattributes` (confirmed to occur identically with `.gitattributes` absent). `git add -A && git diff --cached --stat` shows this conclusively: **only `.gitattributes` has a real content change** (13 insertions, the new file itself) — zero other files have any actual diff.
- [x] Confirmed via four independent methods that no tracked file's content changes as a result of this PR: `git diff`, `git diff --numstat`, `git hash-object` (filtered and `--no-filters`) against the index/HEAD blob hash, and raw byte-level CR/LF counts read via .NET (bypassing any shell text-mode translation).
- [x] `git status --short` on a clean checkout of this branch (no formatter run) shows no unexpected modifications.
- [ ] Full test suite / lint / build — not run; this change touches no source, tests, or build config, only adds a root-level git metadata file with no runtime effect.

- [x] This is the fix for a Windows-specific issue; verified on Windows (the platform where the noise occurs). Linux/macOS CI runners already produce LF, so this is a no-op for them.

## Related Issues

Fixes #253

## Reviewer Notes

- The `binary` overrides list (`*.png`, `*.db`, `*.xlsx`) was derived by tallying all tracked file extensions (`git ls-files` + extension histogram) rather than assumed — these are the only binary-ish assets actually committed to the repo (test fixture images/dbs/excel files). No fonts, icons, or PyInstaller build outputs are currently tracked, so no overrides were needed for those.
- Deliberately did **not** run `git add --renormalize .` — per the issue, that bulk line-ending cleanup of already-committed files is a separate, clearly-labeled follow-up PR if the team wants it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+# Normalize all text files to LF line endings on checkin, regardless of a
+# contributor's local core.autocrlf setting. This matches what Prettier
+# writes by default and what the Linux CI runners already use, avoiding
+# the CRLF/LF "noise" that made git status report hundreds of unmodified
+# files as changed on Windows checkouts (see issue #253).
+* text=auto eol=lf
+
+# Explicit binary overrides for tracked binary assets. git's text=auto
+# heuristic (NUL-byte detection) would likely classify these correctly on
+# its own, but pinning them avoids any risk of accidental normalization.
+*.png binary
+*.db binary
+*.xlsx binary


eberrigan (commented):

## Review Summary

Scaled review note: the repo's standard 5-lens subagent team (code quality/architecture, TDD/testing infra, scientific metadata/UX, security/cross-platform, behavioural edge cases) is built around Electron/React/Python feature changes — IPC handlers, session state, hardware, scan metadata. This PR touches none of that: the diff is a single new `.gitattributes` file (13 lines, 0 deletions), no source, tests, or build config changed. Running all 5 full subagents would have produced mostly "not applicable" noise, so this review was done directly, checking only the dimensions that actually apply to a git-attributes change.

No blocking issues. This is a correct, minimal, appropriately-scoped fix for #253.

## Checks performed

- **Syntax/semantics**: `* text=auto eol=lf` is valid and standard. The three `binary` overrides (`*.png`, `*.db`, `*.xlsx`) are more specific patterns that correctly override the catch-all per git's "last matching pattern wins" attribute precedence — confirmed this is the documented git behavior, not an assumption.
- **Binary coverage completeness**: cross-checked against a full tally of tracked file extensions (`git ls-files` + extension histogram), not assumption. Only `.png` (73 files, sample scan + test-capture images), `.db` (6 SQLite fixtures/dev db), and `.xlsx` (6 Excel test fixtures) are actually-tracked binary-ish assets. No fonts, icons, or PyInstaller build outputs are currently tracked, so no additional overrides are needed — the PR correctly doesn't add speculative entries for asset types that don't exist yet.
- **Scope**: `gh pr view --json files` confirms exactly one file changed, `.gitattributes`, +13/-0. No `git add --renormalize .` was run, matching the issue's explicit instruction to keep any bulk re-normalization as a separate follow-up PR.
- **Verification rigor**: the PR body documents a real gotcha found during testing — after a fresh clone + `npm install` + `npm run format`, `git status --short` initially flags ~460 files as modified, which could easily be mistaken for the bug still existing. The author correctly dug in and confirmed (via `git diff`, `git hash-object` with/without filters, and raw CR/LF byte counts) that this is a pre-existing git-for-windows index stat-cache display quirk, unrelated to this change (reproduces identically without `.gitattributes` present) — and that `git add -A && git diff --cached --stat` shows the true state: only `.gitattributes` itself has any real content change. Good adversarial self-review; this is exactly the kind of check that should happen before trusting `git status` alone as a verification method on Windows.
- **Cross-platform impact**: correctly a no-op for Linux/macOS CI (already LF), and this is the intended Windows-specific fix.

## Suggestions (non-blocking)

- None. The scope is intentionally minimal and matches the issue precisely.

---
*Review by Claude Code (scaled to a single-file config change — full 5-lens subagent team not applicable, see summary)*

PR #252, opened 2026-07-23
chore: standardize .claude/commands dev-command suite

eberrigan:
## Summary

Aligns bloom-desktop's Claude Code `.claude/commands` suite with the lab-wide dev-command standard, using the `standardize-dev-commands` skill's Examine -> Plan -> Apply -> Verify workflow. Every command was checked against this repo's real toolchain (npm + uv, Vitest + pytest, Electron Forge, Prisma, OpenSpec) -- nothing was copied blind from another repo.

## Disposition

| Command | Change |
|---|---|
| `/dev`, `/fix-formatting`, `/test`, `/tdd`, `/run-ci-locally`, `/validate-env` | Added -- were missing entirely |
| `cleanup-branch` -> `/cleanup-merged` | Renamed + hardened: adds a `state == MERGED` gate before any branch deletion and `git fetch --prune`, on top of this repo's existing task-completeness/dependency-order archiving logic |
| `changelog` -> `/update-changelog` | Renamed to the canonical name (no content change needed) |
| `/pr-description` | Updated -- adds the three-state `[x]`/`[!]`/`[ ]` verification convention + an OpenSpec-change section; keeps this repo's detailed TS/Python/Integration/E2E/Hardware/DB checklist |
| `/copilot-review` | Updated -- was a raw GraphQL query dump; now the full triage workflow (categorize by priority, offer-to-address, edge cases) |
| `/pre-merge` | Updated -- fixes references to the two renamed commands, adds the checkbox convention, adds explicit `/fix-formatting` and `/copilot-review` steps |
| `/lint` | Updated (light) -- points to `/fix-formatting` for auto-fix |
| `/new-feature` | Updated -- enriched with explicit `/tdd`/`/lint`/`/test`/`/pre-merge`/`/cleanup-merged` steps; fixed to correctly invoke the `openspec-review` skill (it was written as a slash command, `/openspec-review`, which doesn't exist) |
| `README.md` | Rewritten -- was missing `/ci-debug`, `/docs-review`, and `/copilot-review` entirely, and `/review-pr`'s description didn't match its actual (adversarial 5-subagent) content |
| `/docs-review`, `/ci-debug`, `/coverage` | Kept unchanged -- already excellent, repo-specific, exceed the generic template |
| `/review-pr` | Updated (light) -- fixed the hardcoded repo literal in its GraphQL query (see "Fixes discovered" below); otherwise already excellent, repo-specific, exceeds the generic template |
[...]

PR #191, opened 2026-04-15
refactor: extract GraviScan wiring from main.ts into testable module

eberrigan:
## Summary

- Extract GraviScan wiring logic (state, orchestration, shutdown) from `main.ts` into `src/main/graviscan/wiring.ts` — a side-effect-free module importable in Node test environments
- Add `shutdownGraviScan()` to encapsulate coordinator shutdown + scan log close (required since extracted state is module-private)
- Rewrite `main-wiring.test.ts` to test real exported functions instead of inline reimplementations (25 tests, up from 14)
- Fix pre-existing spec gap: add `rename-error` to Coordinator Event Forwarding events list

Closes #190
Related: #181 (make main process testable), #130 (GraviScan integration epic)

## Details

**Why:** Tests previously re-implemented wiring logic inline because importing `main.ts` triggers Electron side effects. Now tests import and exercise the real production code.

**Known behavior change:** `closeScanLog()` now runs before `closeDatabase()` instead of after. Safe because the scan log is a filesystem write stream with no database dependency (see design.md).

**OpenSpec:** `refactor-extract-graviscan-wiring` — proposal, design, tasks, and spec deltas included.

## Test plan

- [x] 25 unit tests in `main-wiring.test.ts` (14 rewritten + 11 new)
- [x] Full suite: 717 passed, 9 skipped (706 original + 11 new)
- [x] TypeScript compiles clean (`npx tsc --noEmit`)
- [x] Lint clean (`npm run lint`)
- [x] OpenSpec validates (`npx openspec validate --strict`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #256, opened 2026-07-24
fix: skip fetch-scanners E2E tests when live Bloom API creds unavailable

### LINKED ISSUE #254: Flaky/broken E2E test: machine-config-fetch-scanners.e2e.ts scanner fetch timeout (all 3 platforms)
## Problem

`tests/e2e/machine-config-fetch-scanners.e2e.ts` has two tests that consistently time out waiting for the mock scanner fetch to resolve, on all three CI platforms:

- `should populate scanner dropdown after successful fetch` (line ~248)
- `should not require full form completion to fetch scanners` (line ~283)

Both fail with:

\`\`\`
TimeoutError: page.waitForSelector: Timeout 10000ms exceeded.
  - waiting for locator('text=/Found \d+ scanner/') to be visible
\`\`\`

Both tests retry once and fail again with the identical error, then the job reports failure. All other E2E tests in the same run pass (249/251, 248/250, or 249/251 depending on platform — only these two ever fail).

## Where this was observed

Most recently on PR #252 (a docs-only change touching only \`.claude/commands/**\`, which cannot affect Electron/renderer runtime behavior) — `Test - E2E Dev Build` failed identically on **macOS, Ubuntu, and Windows** with the same two tests and the same timeout.

This is not new: the last merged PR (#191, 2026-04-16) also has a failing `Test - E2E Dev Build (macos-latest)` job on its merge commit. CI logs from that run have since expired (7-day retention), so the exact failure signature there couldn't be re-confirmed, but the pattern of this specific job intermittently/consistently failing predates this PR by months.

## Hypothesis

The mock scanner-fetch simulation (whatever backs "Found N scanner(s)" in mock mode) either isn't resolving within the 10s timeout in CI, or a recent-ish change to the mock scanner flow / fetch-scanners UI broke the expected text/timing. Needs investigation into:

- `src/renderer/components/*` fetch-scanners UI flow and the mock scanner backend it talks to
- Whether 10000ms is now too tight for CI runner load, or whether the fetch never actually completes (i.e., real regression vs. flaky timing)

## Suggested next step

Reproduce locally with `npm run test:e2e -- machine-config-fetch-scanners`, capture the Playwright trace (`test-results/**/trace.zip` — already uploaded as CI artifacts on affected runs) to see whether the scanner list ever populates, and check server/mock-hardware logs during the wait window.

eberrigan:
## Summary

Fixes the two E2E tests in `tests/e2e/machine-config-fetch-scanners.e2e.ts` that consistently time out on all 3 CI platforms (#254) by skipping them in CI, where they can never actually succeed.

## Root Cause

These two tests (`should populate scanner dropdown after successful fetch`, `should not require full form completion to fetch scanners`) fill in credentials and click "Fetch Scanners," which triggers a **real network call** (`src/main/config-store.ts` → `supabase.auth.signInWithPassword`) to the live production endpoint `https://api.bloom.salk.edu/proxy`. This isn't mocked — it's a genuine integration test against a real backend.

I confirmed via the actual Playwright trace artifacts from PR #252's CI run (downloaded from GitHub Actions, ubuntu-latest + macos-latest) that the UI shows:

```
⚠️ Authentication failed: fetch failed
```

`fetch failed` is Node's network-layer error (DNS/connection failure), not Supabase's "Invalid login credentials" — so the request never completes; it's not resolving late, it fails outright. Bumping the 10s timeout would not have fixed this.

Supporting evidence this is a pre-existing design gap, not a regression:
- No GitHub Actions secrets are configured for this repo (`gh secret list` is empty), so CI has only ever used the hardcoded fallback credentials in the test file.
- `.env.example` itself documents `BLOOM_TEST_USERNAME`/`PASSWORD`/`BLOOM_ANON_KEY` as being "Used for **manual testing** of Bloom API integration" — this was written for local/manual verification, not unattended public CI runners.
- The issue notes this same job also failed back on PR #191 (April), consistent with it being broken since these tests were added to the CI matrix rather than a recent regression.

The two *other* credential-related tests in the file (loading state, invalid credentials) pass only because their assertions accept `Authentication failed|Failed to fetch` as valid outcomes — they don't require a real successful round-trip.

## Changes

- Add `test.skip(...)` guarding on all three required env vars (`BLOOM_TEST_USERNAME`/`PASSWORD`/`BLOOM_ANON_KEY`), so the two tests are skipped automatically in CI (no secrets configured) but still run for real local/manual verification when a developer sets up `.env.e2e` with real credentials — matching the documented intent of those env vars.
- Replace the hardcoded fallback credentials (a real test account, `bloom_desktop_test@salk.edu`) with generic placeholders consistent with `.env.example`'s style. That account has separately been deleted from the Bloom auth backend.

## OpenSpec Change

No OpenSpec change: this is a test-only fix that restores intended CI behavior (per `openspec/AGENTS.md`, bug fixes and "tests for existing behavior" skip the proposal process). The `fetch-scanners` production capability and its spec (`openspec/specs/ui-management-pages/spec.md`) are unchanged — only how these two tests execute in CI vs. locally.

### Linting

- [x] `npx eslint tests/e2e/machine-config-fetch-scanners.e2e.ts` passes
- [x] `npx tsc --noEmit` reports no errors for this file

### E2E Tests

- [!] Could not run `npm run test:e2e -- machine-config-fetch-scanners` in this dev sandbox — local Electron binary is broken/uninstalled (`Error: Electron failed to install correctly`), a pre-existing environment issue unrelated to this change. Webpack dev build itself compiled successfully.
- [ ] Please confirm locally or via CI that the two tests now report as **skipped** (not failed) and the rest of the suite is unaffected.

### Code Review

This PR went through a 5-agent adversarial review (code quality, testing/TDD, scientific rigor, security/cross-platform, behavioral correctness). No blocking issues found. One real issue surfaced independently by 3 of the 5 agents was fixed in a follow-up commit: the `test.skip()` condition originally only checked `BLOOM_TEST_USERNAME`, not all three required vars, which could produce a confusing local failure instead of a clean skip if only one var was set.

## Related Issues

Fixes #254

## Reviewer Notes

The `test.skip` guard only applies to the 2 tests that need a real network round-trip — the other 5 tests in this file (button visibility/disabled/enabled state, loading state, invalid-credential error) intentionally use fake credentials and don't need this guard, so they still run unconditionally in CI.

### DIFF
--- a/tests/e2e/machine-config-fetch-scanners.e2e.ts
+++ b/tests/e2e/machine-config-fetch-scanners.e2e.ts
@@ -253,7 +253,9 @@ test.describe('Machine Configuration - Fetch Scanners Button', () => {
     // guaranteed path to reach api.bloom.salk.edu. Skip unless a developer
     // has configured real credentials locally. See GH issue #254.
     test.skip(
-      !process.env.BLOOM_TEST_USERNAME,
+      !process.env.BLOOM_TEST_USERNAME ||
+        !process.env.BLOOM_TEST_PASSWORD ||
+        !process.env.BLOOM_ANON_KEY,
       'Requires live Bloom API credentials - set BLOOM_TEST_USERNAME/BLOOM_TEST_PASSWORD/BLOOM_ANON_KEY locally to run this test'
     );
 
@@ -293,7 +295,9 @@ test.describe('Machine Configuration - Fetch Scanners Button', () => {
   test('should not require full form completion to fetch scanners', async () => {
     // Requires a live round-trip to the real Bloom API - see comment above.
     test.skip(
-      !process.env.BLOOM_TEST_USERNAME,
+      !process.env.BLOOM_TEST_USERNAME ||
+        !process.env.BLOOM_TEST_PASSWORD ||
+        !process.env.BLOOM_ANON_KEY,
       'Requires live Bloom API credentials - set BLOOM_TEST_USERNAME/BLOOM_TEST_PASSWORD/BLOOM_ANON_KEY locally to run this test'
     );
 
--- a/tests/e2e/machine-config-fetch-scanners.e2e.ts
+++ b/tests/e2e/machine-config-fetch-scanners.e2e.ts
@@ -246,19 +246,29 @@ test.describe('Machine Configuration - Fetch Scanners Button', () => {
   });
 
   test('should populate scanner dropdown after successful fetch', async () => {
+    // Requires a live round-trip to the real Bloom API with a real test
+    // account. BLOOM_TEST_USERNAME/PASSWORD/ANON_KEY are documented in
+    // .env.example as being for manual testing - they are not (and should
+    // not be) wired up as CI secrets, since GitHub-hosted runners have no
+    // guaranteed path to reach api.bloom.salk.edu. Skip unless a developer
+    // has configured real credentials locally. See GH issue #254.
+    test.skip(
+      !process.env.BLOOM_TEST_USERNAME,
+      'Requires live Bloom API credentials - set BLOOM_TEST_USERNAME/BLOOM_TEST_PASSWORD/BLOOM_ANON_KEY locally to run this test'
+    );
+
     // Fill in VALID test credentials (from .env)
     await window.fill(
       'input[id*="creds-username"]',
-      process.env.BLOOM_TEST_USERNAME || 'bloom_desktop_test@salk.edu'
+      process.env.BLOOM_TEST_USERNAME || 'your-test-username@salk.edu'
     );
     await window.fill(
       'input[id*="creds-password"]',
-      process.env.BLOOM_TEST_PASSWORD || 'bloom_desktop_test_123'
+      process.env.BLOOM_TEST_PASSWORD || 'your-test-password'
     );
     await window.fill(
       'input[id*="creds-anonkey"]',
-      process.env.BLOOM_ANON_KEY ||
-        'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.ewogICAgInJvbGUiOiAiYW5vbiIsCiAgICAiaXNzIjogInN1cGFiYXNlIiwKICAgICJpYXQiOiAxNjg3NTkwMDAwLAogICAgImV4cCI6IDE4NDU0NDI4MDAKfQ.ev7gXAhB8Uv6pgRF9B5oEpmlYI6l15DUIlAQBWSGxPU'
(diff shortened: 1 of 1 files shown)


eberrigan (commented):

## Review Summary

A 5-agent adversarial review (code quality, testing/TDD discipline, scientific rigor & UX, security & cross-platform, behavioral correctness) found **no blocking issues**. The root-cause diagnosis was independently re-verified by two agents that traced past the PR's own claim: one decoded the removed JWT directly, the other traced into `@supabase/auth-js`'s source and confirmed a network-level `fetch()` failure is wrapped into exactly the `"fetch failed"` message string seen in the CI trace artifacts — corroborating that this is a genuine unreachable-network failure, not a slow-but-eventually-successful call or a credentials problem.

## Blocking Issues

None.

## Important Issues (fixed in follow-up commit `2a82833`)

- **Incomplete skip condition** — flagged independently by 3 of 5 agents (code quality, behavioral correctness, testing/TDD). The original `test.skip(!process.env.BLOOM_TEST_USERNAME, ...)` only checked one of the three required env vars. If a developer set `BLOOM_TEST_USERNAME` locally but forgot `BLOOM_TEST_PASSWORD`/`BLOOM_ANON_KEY`, the test would *not* skip — it would run with placeholder password/anon-key values and fail with a confusing 10s timeout instead of a clean skip message. Doesn't affect CI (all three vars are always unset there), but a real local-dev footgun. **Fixed**: now checks all three vars.
- **PR description said "4 tests" unaffected** — actual count is 5 of 7 tests in the describe block (display/disabled/enabled-state + loading-state + invalid-credentials). Cosmetic only; corrected in the PR body.

## Suggestions (not applied — tracked as follow-up, not blocking this fix)

- **Coverage-dark success path** (scientific rigor + testing/TDD agents): once these 2 tests skip in CI, there's no automated signal (nightly job or mock) verifying the "successful fetch populates dropdown" path going forward. Both agents suggest a follow-up issue for either a scheduled workflow with real credentials, or a mocked variant of these 2 scenarios so the success path isn't permanently unverified. Given these tests have *never* had real CI signal since being added (no secrets, no mock — confirmed via `.env.example` and workflow grep), this isn't a regression from this PR, but worth tracking separately from #254.
- Minor duplication between the two skip blocks (near-identical skip message/fill logic) — not worth extracting for just 2 call sites, but flagged in case a 3rd similar test is added later.
- Both skipped tests still pay the full `beforeEach` cost (Electron launch + `prisma db push`) since `test.skip()` inside a test body runs after hooks fire — inherent to Playwright's execution model given only 2 of 7 tests in the file need the skip; not worth restructuring for this fix.

---

Issue #226, opened 2026-05-06
iscan force-install creates permanent apt dependency lock on Ubuntu 24.04


eberrigan:
The Epson iscan package (`iscan 2.30.4-2`) was installed on the GraviScan rig
following the v4 setup docs, which used `sudo dpkg --ignore-depends=libsane -i ...`
to bypass the missing `libsane` package (Ubuntu 24.04 ships `libsane1` instead;
the iscan deb requires the older `libsane (>= 1.0.11-3)`).

This works for scanning — iscan continues to function — but APT now refuses
all subsequent installs/upgrades:

    Unsatisfied dependencies:
     iscan : Depends: libsane (>= 1.0.11-3) but it is not installable

`apt --fix-broken install` would either remove iscan (breaking scanning) or fail.

## Affected rigs

- `graviscan-ms-7c56` (and likely any new rig set up via the v4 build instructions on Ubuntu 24.04)

## Workaround

Install needed packages outside apt — pip/uv for Python deps, local builds for system tools, or `dpkg -i --ignore-depends=libsane <pkg>.deb` for individual debs. We hit this trying to install `libtiff-tools` for the USB card characterization harness; switched to Pillow (Python) instead.

## Possible fixes (unverified, ordered cheapest → biggest)

1. **Check Epson's current Linux drivers.** The iscan we use (2.30.4) is from ~2014. If Epson has released a newer iscan that depends on `libsane1`, the fix is just a re-install. Worth checking https://download.ebz.epson.net/dsc/search/01/search/?OSC=LX for current versions and the [Image Scan! Linux portal](https://download.ebz.epson.net/man/linux/iscan_e.html).

2. **Investigate Windows drivers as a reference / alternative.** Epson Scan 2 for Windows is actively maintained and supports the V600 cleanly. This wouldn't fix the apt lock on the existing Linux rig, but is worth tracking in case:
   - The Linux iscan is truly EOL and we need a forward path
   - We ever consider an OS migration (the v4 build plan's "Option B" called out pyScanLib for cross-platform Python V600 control)
   - Windows drivers expose features we'd want to reach via SANE shims

   (Major lift to actually switch the rig — only worth it if all Linux paths are dead.)

3. **`equivs-build` a fake `libsane` package** satisfying the dep without installing anything. Cheap but creates a phantom dependency that future admins may not recognize.

[...]

PR #257, opened 2026-07-24
docs: real fix for iscan apt dependency lock (#226), not a workaround

eberrigan:
## Summary

Documents and scripts a real, verified fix for the GraviScan scanner-driver apt-lock problem described in #226, replacing the previous `dpkg --ignore-depends` workaround that leaves `apt` permanently broken on any rig it's used on.

## Root cause investigation

Root-caused live against a physically connected Epson Perfection V600 on a fresh Ubuntu 26.04 test box (`pbiob-gh-04`, reachable via Tailscale), verified end-to-end with an actual scan (not just device enumeration):

1. **`libsane` → `libsane1` rename**: already fixed upstream. Modern `iscan` (2.30.6-1) declares `libsane (>= 1.0.11-3) | libsane1` — an OR-dependency satisfied by the renamed package. No workaround needed for this at all; just use a current build. (Confirms issue #226's own "cheapest, unverified" candidate fix.)
2. **`libxml2` → `libxml2-16` rename**: turned out to be *two* separate problems, not one:
   - Package metadata: `apt`/`dpkg` can't resolve `Depends: libxml2` against `libxml2-16`.
   - Runtime linking: even with the metadata satisfied (e.g. via a dummy `equivs` stub), the `epkowa` backend binary is dynamically linked against the old ABI's soname (`libxml2.so.2`), which `libxml2-16` does not provide — a stub alone leaves the backend unable to `dlopen` at all, silently.

   The fix installs the **real** old `libxml2` build from Ubuntu's own archive (`2.12.7+dfsg+really2.9.14`, a security-patched rebuild of the old 2.9.x ABI under a newer version string) — this satisfies both the dependency name and the actual runtime ABI at once. Deliberately not a symlink hack (`libxml2.so.2 -> libxml2-16`'s file) — these are genuinely incompatible ABI versions and a symlink risks silent misbehavior rather than a clean failure.
3. **`epkowa` not enabled** in `/etc/sane.d/dll.conf` by default (it's a non-free, Epson-specific backend, not part of standard `sane-backends`) — installed but never loaded, no error, just no device.
4. **Stale udev permissions** if the scanner was plugged in before the driver's udev rule existed — needs `udevadm control --reload-rules && udevadm trigger` (or a physical replug).

`apt-get check` stays clean after every step — the actual point of doing this properly, since that's exactly what #226 says breaks with the old workaround.

## Changes

- `scripts/setup-graviscan-scanner-driver.sh` — idempotent setup script automating the above (shellcheck-clean, tested twice against the real hardware: fresh install and idempotent re-run)
- `docs/GRAVISCAN_SCANNER_DRIVER_SETUP.md` — full root-cause narrative, quick-start, and troubleshooting, cross-linked from `SCANNER_TESTING.md`/`CAMERA_TESTING.md`

## OpenSpec Change

No OpenSpec change: this is documentation + a setup script, not application behavior (per `openspec/AGENTS.md`, documentation/tooling changes skip the proposal process).

## Testing

- [x] `bash -n scripts/setup-graviscan-scanner-driver.sh` — syntax clean
- [x] `shellcheck` on the Linux test box — zero warnings
- [x] Ran the script for real against `pbiob-gh-04` with a physically connected Epson Perfection V600 — fresh install succeeded, produced a real 3392x4679 RGB TIFF scan
- [x] Re-ran the script a second time to verify idempotency — all steps correctly detected already-done state, exit 0, clean single-line detection output
- [x] `apt-get check` confirmed clean (no broken dependencies) after every step, both runs

## Related Issues

Related to #226 (this is a real fix for fresh installs; the actual production rig `graviscan-ms-7c56` already has the old `--ignore-depends`-broken state and would need a separate, deliberate remediation pass — noted in the doc's troubleshooting section, not attempted here since I don't have access to that machine)

## Reviewer Notes

The two-layer `libxml2` problem (package metadata vs. runtime ABI) is the subtlest part — worth double-checking the reasoning in the doc's "What the script does, and why" section 2 holds up, since it's the one place a shortcut (symlink) was deliberately rejected in favor of the more correct but slightly more involved fix.

### DIFF
--- a/.claude/commands/hardware-testing.md
+++ b/.claude/commands/hardware-testing.md
@@ -300,4 +300,5 @@ Update mocks when:
 - **Camera testing**: `docs/CAMERA_TESTING.md`
 - **DAQ testing**: `docs/DAQ_TESTING.md`
 - **Scanner testing**: `docs/SCANNER_TESTING.md`
+- **GraviScan scanner driver setup**: `docs/GRAVISCAN_SCANNER_DRIVER_SETUP.md`
 - **PyInstaller**: `python/PYINSTALLER.md`
--- a/docs/GRAVISCAN_SCANNER_DRIVER_SETUP.md
+++ b/docs/GRAVISCAN_SCANNER_DRIVER_SETUP.md
@@ -9,7 +9,10 @@ V600 on a fresh Ubuntu 24.04+ machine, without permanently breaking
 Historically, setting up a new GraviScan rig required force-installing
 `iscan` with `sudo dpkg --ignore-depends=libsane -i iscan_*.deb`,
 because Ubuntu 24.04 renamed `libsane` to `libsane1` and the old
-`iscan` build only declared a hard dependency on the former.
+`iscan` build only declared a hard dependency on the former. (Ubuntu
+26.04 separately renamed `libxml2` to `libxml2-16` — a different
+package, a different release, addressed by [step 2](#2-fix-libxml2--a-two-level-problem)
+below. These are two independent renames, not one event.)
 
 That workaround "works" in the sense that scanning functions
 afterward, but it leaves `apt`'s dependency graph in a genuinely
@@ -37,13 +40,23 @@ confirmed with an actual scan (not just device enumeration).
    decision, not an implicit side effect of driver setup).
 2. Extract the downloaded bundle — you should get `install.sh` plus
    `core/`, `data/`, and `plugins/` subdirectories, each with one `.deb`.
-3. Copy it to the target machine and run:
+3. Plug in and power on the scanner (the script's final step performs
+   a real test scan, not just detection, so it needs the device present).
+4. Copy the bundle to the target machine and run:
 
    ```bash
    ./scripts/setup-graviscan-scanner-driver.sh /path/to/extracted/bundle
    ```
 
-4. Verify with an actual scan, not just detection:
+   **Before running**: this script uses `sudo` throughout — it installs
+   system packages (including one fetched from Ubuntu's archive over
+   the network), edits `/etc/sane.d/dll.conf`, and reloads udev rules.
+   Read through it first if you want to know exactly what it touches
+   before granting it root.
+
+   The script's own last step already performs the real-scan
+   verification described below — you shouldn't need to run this
+   manually, but if you want to re-check independently later:
 
    ```bash
    scanimage -d 'epkowa:interpreter:BUS:DEVICE' --resolution 400 --format=tiff > test.tiff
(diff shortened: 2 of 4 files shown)


eberrigan (commented):

## Review Summary

A 3-lens adversarial review (technical accuracy, script safety/robustness, documentation completeness — scaled down from the full 5-agent template since this PR is docs + a shell script, not application code) found no blocking issues, but surfaced real, well-substantiated problems that have since been fixed and re-verified against the physical rig.

## Blocking Issues

None found by any reviewer. (Two issues below were rated blocking by the script-safety reviewer specifically due to `sudo` being involved — fixed regardless of severity label.)

## Fixed in follow-up commit `3f494a1`

- **Unverified network install as root** (script-safety + doc-completeness agents, independently): the `libxml2` compatibility download used plain `http://`, no `-f` flag (so a 404 error page would be silently saved and fed to `dpkg`), and no integrity check before `sudo dpkg -i`. Now: `https://`, `curl -fSL`, and a pinned SHA256 checksum verified with `sha256sum -c` before install. **Verified, not just written**: tested both the correct checksum (passes) and a deliberately wrong one (fails with exit 1, which `set -e` catches) directly on the rig.
- **Unsound idempotency check** (script-safety agent): the libxml2 fix-detection logic couldn't distinguish "already fixed" from "the backend .so is missing entirely because Step 1 silently failed" — both produced the same "nothing to do" message. Now explicitly distinguishes missing-file / unresolved / already-fixed states, and uses `LC_ALL=C` so the check isn't locale-dependent.
- **Script's own claim didn't match its verification** (technical-accuracy agent): the doc says "confirmed with an actual scan," but the script's own Step 6 only checked device enumeration. Step 6 now performs a real low-resolution test scan and validates the TIFF output — re-run against the physical V600 on `pbiob-gh-04` to confirm.
- **Rename attribution error** (technical-accuracy agent, verified via Launchpad): the doc implied `libsane`→`libsane1` and `libxml2`→`libxml2-16` happened in the same release (24.04). They're two releases apart — 24.04 and 26.04 respectively. Corrected.
- **Missing troubleshooting entry** (doc-completeness agent): the script had a comment explaining what to do if the pinned archive package is ever removed, but the doc's Troubleshooting section didn't. Added.
- **Discoverability gap** (doc-completeness agent): the new doc wasn't cross-linked from `.claude/commands/hardware-testing.md` or `docs/SCANNER_TESTING.md`. Fixed both directions.
- **No sudo/security framing before Quick Start** (doc-completeness agent): added an explicit note on what the script touches (system packages, a network fetch, SANE config, udev) before telling readers to run it.

## Suggestions Noted, Not Applied

- Pinning the old libxml2 build's own transitive dependencies (e.g. ICU) wasn't checked before install — `set -e` aborts cleanly if that ever fails; deemed acceptable given this is a narrow, well-scoped compatibility package, not core system infrastructure.

---
*Review by Claude Code (Technical Accuracy · Script Safety & Robustness · Documentation Completeness — scaled to this PR's actual risk surface rather than the full 5-lens template)*

PR #258, opened 2026-07-24
fix: add --scan-worker mode to main.py for packaged builds (#206)

### LINKED ISSUE #206: bloom-hardware: unrecognized arguments '-m graviscan.scan_worker --scanner-id ... --mock'
## Problem

Starting a scan in GraviScan mock mode fails immediately with:

\`\`\`
[ScanWorker:<scanner-id>] usage: bloom-hardware [-h] [--ipc]
[ScanWorker:<scanner-id>] bloom-hardware: error: unrecognized arguments: -m graviscan.scan_worker --scanner-id <id> --mock
[ScannerSubprocess:<scanner-id>] Exited: code=2, signal=null
[ScanCoordinator] Subprocess <scanner-id> exited with code 2
\`\`\`

The renderer error: \`Scanner <scanner-id> process exited (code 2) before becoming ready\`.

## Root cause hypothesis

The \`ScannerSubprocess\` spawn invokes:
\`\`\`
/path/to/dist/bloom-hardware -m graviscan.scan_worker --scanner-id <id> --mock
\`\`\`

But the bundled \`bloom-hardware\` PyInstaller binary's argparse only accepts \`[-h] [--ipc]\` — it does NOT support the \`-m <module>\` flag (which is a CPython interpreter flag, not a script-level flag).

In dev mode with \`uv run python -m graviscan.scan_worker\`, the \`-m\` is interpreted by Python itself. In bundled mode, the binary IS the interpreter and the script combined; \`-m\` becomes a positional arg that argparse rejects.

This is a packaging mismatch between dev-mode invocation and bundled-mode invocation. Likely affects all GraviScan scans in packaged builds and possibly in dev too if the spawn code points at \`dist/bloom-hardware\`.

## Reproduction

1. \`GRAVISCAN_MOCK=true npm run start\`
2. Configure scanners, set up Metadata, click Start Scan
3. Observe terminal log

## Proposed fix (need investigation)

Option A: extend \`bloom-hardware\` argparse to accept \`-m <module>\` and dispatch to that module's \`main()\`. This requires changes to the Python entry point in \`hardware/bloom_hardware/__main__.py\` (or wherever the binary is built from).

Option B: if \`bloom-hardware\` already has multiple subcommands (e.g., \`--ipc\` for cylinder), add a new subcommand \`--graviscan-scan-worker\` that maps to the same code path.

Option C: ship a separate \`graviscan-scan-worker\` binary and have \`ScannerSubprocess\` invoke that directly (no \`-m\`).

The right answer depends on the existing binary structure — needs a quick investigation before deciding.

## Acceptance criteria

- [ ] Scan starts successfully in \`GRAVISCAN_MOCK=true\` mode without subprocess argparse errors
- [ ] Test in both dev mode (\`npm run start\`) and packaged mode (\`npm run package\`)
- [ ] Existing CylinderScan invocations still work

## Related

- Likely pre-existing; not introduced by \`fix-scanner-config-save-flow\` (PR #196), but surfaced during its smoke test
- Tests \`graviscan-scanning.e2e.ts\` may also be hitting this — worth verifying CI status

eberrigan:
## Summary

Increment 1 of the [GraviScan backend-hardening incremental port plan](docs/superpowers/plans/2026-07-24-graviscan-backend-hardening.md). Fixes issue #206: `main`'s `python/main.py` had no support for the `--scan-worker` flag, but `src/main/graviscan/scanner-subprocess.ts` already spawns `bloom-hardware --scan-worker --scanner-id ... --device ...` in packaged builds — so **GraviScan currently cannot run scan workers in any packaged build**, and per #206's own repro steps, this also fails in `GRAVISCAN_MOCK=true npm run start` dev mode.

## Root Cause

Confirmed via `git show main:python/main.py | grep scan-worker` (no matches) — the CLI entry point simply had no routing for the flag Electron already spawns with.

## Changes

- `python/graviscan/scan_worker.py` — extracted `run_worker(scanner_id, device, mock)` out of `main()`'s CLI body as a pure refactor, so it's a shared entry point for both the existing dev CLI (`python -m graviscan.scan_worker`) and the new packaged-build path.
- `python/main.py` — added `scan_worker_mode()` (mirrors the existing `ipc_mode` import-fallback pattern) plus `--scan-worker`/`--scanner-id`/`--device`/`--mock` argparse flags and routing.
- Fixed a validation gap the port would otherwise have inherited from the source commit: `--device` is now only required when `--mock` is not set (mirroring `scan_worker.py`'s own `main()` semantics), because `scanner-subprocess.ts` omits `--device` entirely for packaged+mock spawns — the original snippet would have left packaged+mock builds broken by a different error than #206.
- `python/tests/test_main.py` — 5 new CLI-routing tests (real success, mock without `--device`, missing `--scanner-id`, missing `--device`), mocking only `run_worker` itself so the real import-fallback/routing logic executes.

## OpenSpec Change

None — this is a narrow, pre-declared bug-fix restoring already-intended behavior (issue #206), which the plan's Global Constraints and `openspec/AGENTS.md` both say doesn't need a formal proposal.

### Python
- [x] `uv run pytest python/tests/test_main.py -v` — 14/14 passing
- [x] `uv run pytest python/tests/test_scan_worker.py -v` — 55/56 passing (1 pre-existing failure, `TestUSBResetPathConstruction::test_path_from_device_name`, patches the Linux-only `fcntl` module which doesn't exist on Windows dev machines — confirmed present on `main` before this change, unrelated to this diff)
- [x] `uv run ruff format --check python/main.py` — compliant

### TypeScript
- [x] `npx tsc --noEmit` — no errors
- [!] `npm run lint` — pre-existing worktree-environment issue (duplicate `eslint-plugin-import` resolution between nested `node_modules`), unrelated to this change; not reproducible on a normal `main` checkout

### Physical rig verification (pbiob-gh-04, test box — not the production rig)
- [x] Reproduced the exact pre-fix bug: `python3 python/main.py --scan-worker --scanner-id test-scanner --device mock-device --mock` → `error: unrecognized arguments: --scan-worker ...` (exit 2)
- [x] Confirmed the fix resolves it: same command now runs cleanly (mock init → ready event → clean shutdown, exit 0)
- [x] Confirmed the packaged+mock spawn form specifically (`--scan-worker --scanner-id test-scanner --mock`, no `--device`, matching what `scanner-subprocess.ts` actually sends): also runs cleanly post-fix

Node/npm are not installed on this test rig, so verification used direct Python CLI invocation (the exact command Electron spawns) rather than `GRAVISCAN_MOCK=true npm run start`.

### Code Review

Went through subagent-driven-development: a task-level review found the CLI-routing tests initially mocked one layer too high (`scan_worker_mode` instead of `run_worker`), fixed and re-verified. A final whole-branch review (on the most capable available model) found the packaged+mock validation gap described above — this deviates from the plan's literal Step 4 snippet (and the original source commit), confirmed with the repo owner before fixing, then fixed and re-verified clean.

## Related Issues

Fixes #206

### DIFF
--- a/python/graviscan/scan_worker.py
+++ b/python/graviscan/scan_worker.py
@@ -802,6 +802,25 @@ class ScanWorker:
         log(self.scanner_id, "Done")
 
 
+def run_worker(scanner_id: str, device: str, mock: bool = False) -> None:
+    """Entry point for scan worker — used by both dev mode and production PyInstaller.
+
+    Creates a ScanWorker, initializes SANE, and enters the command loop.
+    Exits with code 1 if initialization fails, 0 on clean exit.
+    """
+    worker = ScanWorker(
+        scanner_id=scanner_id,
+        device_name=device,
+        mock=mock,
+    )
+
+    if not worker.initialize():
+        sys.exit(1)
+
+    worker.run()
+    sys.exit(0)
+
+
 def main():
     parser = argparse.ArgumentParser(
         description="GraviScan worker for a single scanner"
@@ -820,17 +839,7 @@ def main():
     if not args.mock and not args.device:
         parser.error("--device is required unless --mock is specified")
 
-    worker = ScanWorker(
-        scanner_id=args.scanner_id,
-        device_name=args.device or "mock-device",
-        mock=args.mock,
-    )
-
-    if not worker.initialize():
-        sys.exit(1)
-
-    worker.run()
-    sys.exit(0)
+    run_worker(args.scanner_id, args.device or "mock-device", args.mock)
 
 
 if __name__ == "__main__":
--- a/python/main.py
+++ b/python/main.py
@@ -74,15 +74,34 @@ def ipc_mode():
     run_ipc_loop()
 
 
+def scan_worker_mode(scanner_id: str, device: str, mock: bool = False):
+    """Run as a scan worker subprocess for a single scanner."""
+    try:
+        from python.graviscan.scan_worker import run_worker
+    except ModuleNotFoundError:
+        from graviscan.scan_worker import run_worker  # type: ignore[import-not-found]
+    run_worker(scanner_id, device, mock)
+
+
 def main():
-    """Main entry point - routes to interactive or IPC mode."""
+    """Main entry point - routes to interactive, IPC, or scan-worker mode."""
     parser = argparse.ArgumentParser(description="Bloom Hardware Interface")
     parser.add_argument(
         "--ipc", action="store_true", help="Run in IPC mode for Electron communication"
     )
+    parser.add_argument(
+        "--scan-worker", action="store_true", help="Run as scan worker subprocess"
+    )
+    parser.add_argument("--scanner-id", type=str, help="Scanner UUID (scan-worker mode)")
+    parser.add_argument("--device", type=str, help="SANE device name (scan-worker mode)")
+    parser.add_argument("--mock", action="store_true", help="Use mock scanner (scan-worker mode)")
     args = parser.parse_args()
 
-    if args.ipc:
+    if args.scan_worker:
+        if not args.scanner_id or not args.device:
+            parser.error("--scan-worker requires --scanner-id and --device")
+        scan_worker_mode(args.scanner_id, args.device, args.mock)
+    elif args.ipc:
         ipc_mode()
     else:
         interactive_mode()
--- a/python/main.py
+++ b/python/main.py
@@ -86,10 +86,11 @@ def scan_worker_mode(scanner_id: str, device: str, mock: bool = False):
 def main():
     """Main entry point - routes to interactive, IPC, or scan-worker mode."""
     parser = argparse.ArgumentParser(description="Bloom Hardware Interface")
-    parser.add_argument(
+    mode_group = parser.add_mutually_exclusive_group()
+    mode_group.add_argument(
         "--ipc", action="store_true", help="Run in IPC mode for Electron communication"
     )
-    parser.add_argument(
+    mode_group.add_argument(
         "--scan-worker", action="store_true", help="Run as scan worker subprocess"
     )
     parser.add_argument(
--- a/python/main.py
+++ b/python/main.py
@@ -79,7 +79,7 @@ def scan_worker_mode(scanner_id: str, device: str, mock: bool = False):
     try:
         from python.graviscan.scan_worker import run_worker
     except ModuleNotFoundError:
-        from graviscan.scan_worker import run_worker  # type: ignore[import-not-found]
+        from graviscan.scan_worker import run_worker  # type: ignore[import-not-found,no-redef]
     run_worker(scanner_id, device, mock)
 
 
--- a/python/main.py
+++ b/python/main.py
@@ -92,15 +92,23 @@ def main():
     parser.add_argument(
         "--scan-worker", action="store_true", help="Run as scan worker subprocess"
     )
-    parser.add_argument("--scanner-id", type=str, help="Scanner UUID (scan-worker mode)")
-    parser.add_argument("--device", type=str, help="SANE device name (scan-worker mode)")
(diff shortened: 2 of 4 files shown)


eberrigan (commented):

## Review Summary

5-agent adversarial review (Code Quality · Testing/TDD · Scientific Rigor · Security/Cross-Platform · Behavioural Correctness) of the full branch (`87f0fdc..958cd3f`). **No blocking issues.** The core fix is correct, behavior-preserving for the existing dev CLI, matches exactly what `scanner-subprocess.ts` actually spawns, and is well-tested for the paths that matter in production. CI is green (Lint-Python/mypy, Test-Python, TypeScript, Node lint, all Integration jobs, Packaged-App-DB); only 3 unrelated E2E Playwright jobs remain pending.

One note on process: the code-quality and security-review subagents each independently mutated `python/main.py` mid-review (stripping/altering the `# type: ignore` comments while investigating the mypy suppression) despite being read-only reviewers. Both mutations were caught and reverted before anything was committed/pushed — flagging for transparency, not as a PR defect.

## Important Issues (non-blocking, worth a decision)

1. **Three new test docstrings overclaim coverage of the PyInstaller-fallback branch.** `test_main.py`'s docstrings say the tests are structured "so `scan_worker_mode`'s own try/except import-fallback logic actually executes" — but coverage data (`main.py` lines 81-82) confirms only the `try` branch ever runs; the `except ModuleNotFoundError` fallback (the actual PyInstaller-bundle code path) is untested in both dev and CI. This mirrors a pre-existing, equally-untested gap in `ipc_mode()` — not a regression — but the docstring wording should be corrected to not overstate what's verified. (Flagged independently by 2 reviewers.)

2. **No test locks in `run_worker()`'s exit-code contract.** Neither `sys.exit(1)` on init failure nor `sys.exit(0)` on clean completion has direct test coverage — `test_main.py` mocks `run_worker` entirely, and `test_scan_worker.py` has no `run_worker`-specific test. Pre-existing gap (same as the old inline `main()` body before extraction), but `run_worker` is now a shared, production-facing entry point for the first time, so it's a good moment to close it. (Flagged independently by 2 reviewers.)

3. **No mutual-exclusion guard between `--scan-worker` and `--ipc`.** `if args.scan_worker: ... elif args.ipc: ...` silently prioritizes `--scan-worker` if both are ever passed together — not reachable via any current caller (`scanner-subprocess.ts` only ever sends one or the other), but nothing stops a future caller from doing it silently wrong. `argparse.add_mutually_exclusive_group()` would make this loud instead of silent. (Flagged independently by 2 reviewers.)

## Suggestions

- Add a one-line comment on `scan_worker_mode`'s `# type: ignore[import-not-found,no-redef]` explaining *why* it needs an extra error code beyond `ipc_mode`'s single-code suppression (verified: mypy treats `graviscan.scan_worker` and `python.graviscan.scan_worker` as distinct despite resolving to the same file — a known false-positive class for this dual-import-path idiom, not a masked bug).
- `"mock-device"` placeholder string is now duplicated across `scan_worker.py` and `main.py` with no shared constant — low risk today (verified `ScanWorker.initialize()` never touches `device_name` when `mock=True`), but worth a named constant to prevent drift.
- `python/main.py`'s module-level docstring (lines 2-9) still says "Can run in two modes: IPC / Interactive" — now factually stale, missed when `main()`'s own docstring was updated for the new scan-worker mode.
- The `--device`-required-unless-`--mock` validation is now duplicated verbatim (with slightly different wording) across `main.py` and `scan_worker.py`'s two argparse parsers — worth a shared helper if a third entry point is ever added.

## Assessment

All 5 lenses agree: correct, faithful port with a genuine bug fix (the packaged+mock validation gap), no security/cross-platform/data-integrity risk, CI green. The Important items above are real but scoped to test-coverage completeness and defensive robustness for currently-unreachable code paths — reasonable to fix now (cheap) or track as fast-follow, your call.

---

Issue #245, opened 2026-05-27
Add UI banner showing Slack webhook + LIBUSB_ENDPOINT_RECOVERY env-var state


eberrigan:
## Context

During the PR #237 rig validation brainstorm on 2026-05-26, the
operator asked: "How do I know if the env was set up correctly?"
A code search confirmed there is no UI indicator — operators have
to inspect journald (or grep the startup log) for these lines:

\`\`\`
[GraviScan] Slack webhook URL loaded from ~/.bloom/.env
   OR
[GraviScan] BLOOM_GRAVISCAN_SLACK_WEBHOOK_URL not set — Slack notifications disabled

[GraviScan] LIBUSB_ENDPOINT_RECOVERY explicitly set to: true
   OR
[GraviScan] LIBUSB_ENDPOINT_RECOVERY not set in env — wrapper defaults to ON
\`\`\`

## Why this matters

If the Slack webhook is misconfigured (URL missing, wrong, or
rate-limited at the Slack end), the operator gets NO wedge alerts —
which means a wedged scanner can produce missing-data gaps for
hours or days before anyone notices. The whole point of the
WedgeDetector + SlackNotifier infrastructure (#236) is real-time
notification; silent misconfiguration defeats that.

LIBUSB_ENDPOINT_RECOVERY is less critical (defaults to ON in the
C-shim when the env var is absent, per
\`src/main/native/libusb-filter.c:140\`), but the operator should
still be able to verify it explicitly.

## Proposed UI

Add a small status pill in the app header (top-right corner) that
shows the env-var state at-a-glance:

\`\`\`
 [✓ Slack: on]   [✓ libusb-recovery: on]
\`\`\`

Colors:
- Green ✓ — both configured + Slack reachable (last test < 1 hour ago)
- Amber ⚠ — configured but stale-test (> 1 hour) or one missing
- Red ✗ — Slack URL absent or last test failed

Click the pill → modal with details + 'Test Slack' button that
POSTs a no-op message to the webhook + reports the HTTP status.

## Why a pill, not a banner

A banner is wasted screen space when both are green (the common
case). A pill is unobtrusive when green, attention-grabbing when
red/amber. Pattern matches the existing Mock Mode banner (which
only appears when active).

## Code locations

- Env-var load: \`src/main/main.ts:1251-1295\`
- Env-var hydration: \`src/main/main.ts:1260-1280\`
[...]

Issue #232, opened 2026-05-13
V600 only supports [400/800/1600/3200] dpi; requesting 1200 silently rounds to 1600 but DB records 1200


eberrigan:
## Observed on branch `feature/graviscan-prod`

(Issue surfaced 2026-05-12 while preparing the T1 card-characterization test — related to #228.)

## Symptom

The Configure Scanner UI offers a resolution dropdown including **1200 dpi labelled "(Recommended)"** ([`src/renderer/ConfigureScanner.tsx:374`](../blob/feature/graviscan-prod/src/renderer/ConfigureScanner.tsx#L374) and [`src/renderer/components/graviscan/ScannerConfigSection.tsx:588`](../blob/feature/graviscan-prod/src/renderer/components/graviscan/ScannerConfigSection.tsx#L588)). The `GraviConfig.resolution` schema default is 1200, and every persisted `GraviScan.resolution` row in the production DB records 1200.

**But the V600's actual SANE driver reports only `[400, 800, 1600, 3200]` as supported resolutions** (full bloom log line below). When the app requests 1200, the SANE driver rounds up — almost certainly to **1600** — and silently scans at that resolution. The DB still records 1200.

## Evidence

From today's running bloom-graviscan log (any of the 5 scanner subprocesses logs the same thing on startup):

```
2026-05-12T22:35:41.612Z  [4e23d765-...] Resolution option: (12, 'resolution', 'Scan resolution',
  'Sets the resolution of the scanned image.', 1, 4, 4, 5, [400, 800, 1600, 3200])
```

This is the SANE_CONSTRAINT_WORD_LIST descriptor from epkowa — the V600 device-driver allows only those four discrete values, nothing in between.

Confirmation from a standalone `scanimage` invocation at 300 dpi:

```
scanimage: rounded value of resolution from 300 to 400
```

…and the resulting 50×50 mm TIFF is exactly the byte-size expected for a 400-dpi 8-bit RGB scan (787×787×3 ≈ 1.86 MB matches the 1,851,264 B output).

So 300→400 confirmed empirically. The same rounding logic applied to 1200 lands on 1600 (or 800, depending on tie-break; the bloom log line and TIFF bit-rate would settle it but the point stands either way: it's not 1200).

## Why this matters

Two distinct problems, neither cosmetic:

[...]

Issue #231, opened 2026-05-13
Per-scanner grid_mode silently fails to save (UPDATE handler omits the field) on feature/graviscan-prod


eberrigan:
## Observed on branch `feature/graviscan-prod`

(Issue surfaced 2026-05-12 while preparing the T1 card-characterization test — related to #228 and #230.)

## Symptom

On the Configure Scanner page, changing a scanner's grid-mode dropdown from 4-Grid to 2-Grid appears to save successfully — the main-process log even reports:

```
[GraviScan:SAVE] Attempting to save scanners to database: [ ... grid_mode: "2grid", ... ]
[GraviScan:SAVE] Updated scanner: { id: ..., name: 'Scanner 1', usb_bus: null, usb_device: null }
[GraviScan:SAVE] Successfully saved 5 scanners to database
```

…but querying `GraviScanner` after the save shows `grid_mode='4grid'` everywhere. The `updatedAt` field advances on each save (because Prisma `@updatedAt` triggers on any UPDATE), but `grid_mode` itself never changes.

## Root cause

[`src/main/graviscan-handlers.ts`](../blob/feature/graviscan-prod/src/main/graviscan-handlers.ts) `graviscan:save-scanners-db` (handler starting around L436). The UPDATE `data` block is:

```ts
const updated = await db.graviScanner.update({
  where: { id: existing.id },
  data: {
    name: scanner.name,
    display_name: scanner.display_name ?? existing.display_name ?? null,
    vendor_id: scanner.vendor_id,
    product_id: scanner.product_id,
    usb_port: scanner.usb_port || null,
    usb_bus: scanner.usb_bus || null,
    usb_device: scanner.usb_device || null,
  },        // ← grid_mode is NOT in this data block
});
```

The CREATE block (same handler, a few lines below) is similar — also no `grid_mode`. So new rows take the Prisma schema default `'4grid'`, and updates never touch `grid_mode` regardless of what the renderer sends.

The handler's TypeScript input type does include `grid_mode?: string`, and the renderer ([`ConfigureScanner.tsx`](../blob/feature/graviscan-prod/src/renderer/ConfigureScanner.tsx) `handleGridModeChange`) does populate it correctly in the payload — the data just gets silently dropped in the Prisma call.

## Why this is a more serious bug than it looks

[...]

Issue #230, opened 2026-05-12
Stale GraviScanner rows persist after USB reconfiguration; no UI to disable/remove


eberrigan:
## Observed on branch `feature/graviscan-prod`

(Issue surfaced 2026-05-12 while preparing the T1 card-characterization test — context related to #228.)

## Symptom

After reconfiguring USB cables (moving scanners to different host ports — e.g., from PCIe expansion cards to motherboard rear-panel jacks), the Configure Scanner page shows **all historically-detected scanners**, not just the currently-plugged ones. In our case the page showed **13 scanners** when only **5** were physically connected.

The `validateConfig` flow then permanently reports `status='mismatch'` with messages like:

> Scanner configuration has changed. Missing: Scanner 4, Scanner 1, Scanner 2, … . New scanners on ports: 1-5, 1-6, 3-1, 5-1.

Result: the operator can't reach a `valid` configuration state — the test was blocked.

## Root cause

[`src/renderer/ConfigureScanner.tsx`](../blob/feature/graviscan-prod/src/renderer/ConfigureScanner.tsx) `handleDetect()` upserts each detected scanner by `usb_port` via `saveScannersDb`, but it **never deletes or disables** rows whose `usb_port` is no longer in the detection set.

[`src/main/graviscan-handlers.ts`](../blob/feature/graviscan-prod/src/main/graviscan-handlers.ts) `graviscan:validate-config` (≈ L825) and `graviscan:get-scanner-status` (≈ L988) both filter \`where: { enabled: true }\`. So if stale rows had \`enabled=false\` they'd vanish from the UI — but **nothing in the UI sets that flag**, and there's no per-row enable/disable toggle on Configure Scanner.

[`src/renderer/hooks/useScannerConfig.ts`](../blob/feature/graviscan-prod/src/renderer/hooks/useScannerConfig.ts) (\`validateScannerConfig\`, around L213-290) then computes \`matched / missing / new\` against the persisted enabled set and returns \`mismatch\` whenever stale enabled rows exist that no longer enumerate.

\`?reset=true\` ([ConfigureScanner.tsx:103-108](../blob/feature/graviscan-prod/src/renderer/ConfigureScanner.tsx#L103-L108)) only clears React state, not DB. \`graviscan:reset-scanners\` kills scanner subprocesses and shuts down the coordinator but doesn't touch the DB either. So there is **no in-app path** for an operator to clear stale rows on this branch.

[...]

PR #273, opened 2026-07-30
GraviScan: add Configure Scanner UI (roadmap Tier 1)

eberrigan:
## Summary

Tier 1 of the [GraviScan renderer roadmap](docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md) (5-tier plan to close the gap left by the backend-parity port in PRs #267-272: `main` had zero GraviScan renderer UI). This PR:

- Adds `src/renderer/ConfigureScanner.tsx` — the GraviScan hardware-setup page (detect/save/remove scanners, global resolution + grid-mode config, USB reset), mode-gated at `/configure-scanner`.
- Closes the last preload-wiring gap from the backend port: `getScannerStatus` (backend handler already existed and was tested, just unreachable from the renderer).
- Implements two already-accepted-but-unbuilt OpenSpec requirements: **DPI Dropdown Restricted to Validated Set** (trims `GRAVISCAN_RESOLUTIONS` to `[200,400,600,800,1200,1600]`, closing #232's UI half) and **Per-Scanner Remove Button** (closing #230).
- Adds a small env-status banner for #245 (Slack webhook / libusb-recovery configured indicators — booleans only, never the webhook URL).

This also bundles the roadmap doc and both Tier 1/2 OpenSpec proposals (`openspec/changes/add-graviscan-configure-scanner-ui/`, `openspec/changes/add-graviscan-data-layer-and-events/`), since this branch descended from that shared commit before diverging. **Tier 2's implementation is a separate PR** (`add-graviscan-data-layer-and-events`) and doesn't depend on this one merging first — see the roadmap doc's coordination note on the one file they both touch (`preload.ts`, disjoint sections).

Both OpenSpec proposals went through a 5-agent adversarial `openspec-review` pass; every BLOCKING/IMPORTANT finding was fixed before implementation (see each proposal's `design.md` Decisions/Risks sections).

## Notable implementation decisions (see `design.md` for full detail)

- **Global scan-active gate, not per-row status**, for the Remove-button and Reset-USB safety guards. The real `getScannerStatus()` API can't distinguish "initializing" from "actively scanning" per-row (both report `starting`), so the original accepted spec's per-row `scanning` check was unimplementable as written — this was caught during adversarial review and fixed to gate on the session-level `getScanStatus().isActive` instead, which is both honest about the real architecture (the coordinator is a process-wide singleton) and actually enforceable.
- **Reset USB is blocked while a scan is active** and labeled "Reset All USB Connections" with explicit blast-radius copy — the original draft had no guard at all on this, the single most consequential action on the page.
- **One global grid-mode control**, not per-scanner — `main`'s `GraviScanner` model has no per-scanner `grid_mode` column (confirmed; that gap is issue #231, assigned to Tier 2), so building per-row UI against it would silently no-op.
- Found and fixed a **pre-existing type/runtime mismatch**: `electron.d.ts` declared `resetUsb()` as returning `ResetUsbResult` directly, but it's actually wrapped via `wrapHandler`'s `{success, data}` envelope at runtime. No prior renderer code called it, so nothing had caught this until this PR's first real consumer.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npm run test:unit` — 1092 passed, same 7 pre-existing failures as `main`'s baseline (Windows path-separator assertions, unrelated to this change), verified by exact test name before and throughout implementation — zero regressions
- [ ] **E2E tests** (3 new cases in `tests/e2e/graviscan-ipc.e2e.ts`: Detect+resolution-options, Remove flow, Reset USB) — written, `tsc`/lint-clean, follow this file's established conventions, but **could not be executed** in the sandboxed environment this PR was developed in (`npm run start`'s Electron window launches and exits silently, code 0, no error — an unresolved environment limitation, not a code issue). **Needs a real run in CI or on a machine that can launch Electron before merging.**
- [ ] Manual verification with real (or `GRAVISCAN_MOCK=true`) hardware

## OpenSpec

- Proposal: `openspec/changes/add-graviscan-configure-scanner-ui/`
- `npx openspec validate add-graviscan-configure-scanner-ui --strict` passes
- `tasks.md`: 100% complete except the E2E-execution checkbox (left honestly unchecked, see Test plan above)


eberrigan (commented):

## Review Summary

Solid, well-tested page component (25 unit tests, 3 real E2E tests, a genuinely-fixed CI-caught race condition) with real design-doc rigor. But the 5-lens adversarial pass surfaced **3 BLOCKING issues** — all real, concrete, reachable-in-normal-operation bugs, not hypotheticals — plus several IMPORTANT gaps worth closing before merge.

## Blocking Issues

1. **`scanActive` gate is stale-by-construction, not just for out-of-band callers.** `handleResetUsb`/`handleRemove` gate on a React state variable that is only refreshed on mount, inside the two action handlers themselves, and a polling loop that **only arms if a row is already `starting` at the time the effect runs**. There's no subscription to the existing `graviscan:scan-event`/`onScanEvent` push channel (already wired in `preload.ts`). A technician with this page open while the rig is idle, who then starts a scan from elsewhere (or the page just sits on a stale idle snapshot), gets no state update — **Reset All USB Connections** and **Remove** stay clickable and will fire against a live experiment. This undercuts the design.md framing that the gap is limited to hypothetical out-of-band IPC callers — it's reachable through this page's own normal idle→scan lifecycle. Fix: re-check `getScanStatus()` synchronously before firing either action, and/or subscribe to `onScanEvent` to invalidate `scanActive` immediately on any scan start rather than relying on row-status polling as a proxy.

2. **Detect and Reset USB have no mutual exclusion — a plain double-click-speed sequence (click Detect, then click Reset while Detect is still running) reproduces the same coordinator double-init race class that CI just caught and this PR fixed.** Neither button is disabled while the *other* operation is in flight (only self-disabled via `isDetecting`/`isResettingUsb`). `saveScannersToDB`'s post-save spawn loop and `resetUsb()`'s own shutdown→re-detect→re-init cycle both mutate `ScanCoordinator`'s subprocess map concurrently, and the `pendingAdds` dedup in `addScanner()` only protects the already-scanning case, not this one. This is a *different*, still-open instance of the exact bug class the Reset-USB fix in this PR addresses — recommend disabling Detect while `isResettingUsb` and vice versa, ideally with a shared lock at the IPC/coordinator layer too since a non-UI caller of either channel remains unguarded.

3. **The Save success banner ("Configuration saved successfully!") makes no mention that resolution/grid-mode changes don't currently affect any real scan.** Confirmed: `startScan()` sources `resolution`/`grid_mode` from caller-supplied params, not from the saved `GraviConfig`. An operator who "fixes" resolution here and sees an unqualified green success message has a reasonable, false belief the next scan uses the new value — this is exactly the kind of silent metadata/provenance gap that matters on a live rig. This is documented only in `design.md`'s Risks section, which a lab tech will never read. Needs an in-app qualifier near Save (e.g. "Not yet applied to new scans automatically in this build").

## Important Issues

- **Type-safety gap left half-closed**: only `resetUsb`/`getScannerStatus` got real return types in `electron.d.ts`; `detectScanners`, `getConfig`, `saveConfig`, `saveScannersToDB`, `getScanStatus` are still `Promise<any>` — the exact kind of unenforced shape mismatch that let the just-fixed `resetUsb()` bug ship silently. Recommend typing all five now that the correct wrapped shapes are known.
- **No error path on the initial-load `useEffect`'s `Promise.all`** — if any of the four mount-time IPC calls ever rejects, `setLoading(false)` is never reached and the page hangs on "Loading configuration..." forever with no diagnostic. `handleDetect`/`handleResetUsb` use `try {} finally {}` with no `catch`, same silent-failure risk if `ipcRenderer.invoke()` ever actually rejects.
- **Neither `handleDetect` nor `handleResetUsb` has a function-body reentrancy guard** — both rely solely on the `disabled` prop, which doesn't commit until after a render flush. A rapid physical double-click or auto-repeating key can fire both before the first commits.
- **No unmount/cancellation guard in any of the four action handlers** (unlike the mount effect, which has one). Navigating away mid-`detectScanners()` and back, then clicking Detect again, reproduces the same overlapping-coordinator-operations race as Blocking #2 — via navigation instead of a double-click.
- **Reset-USB unit test guards the fix's symptom, not the causal invariant** — it asserts `detectScanners` wasn't called, but not `saveScannersToDB`. The real regression guard is the E2E test (real subprocess timing); the unit test is fix-documentation, not a race-condition guard, and the PR's own commit message concurs.
- **No test for double-click/rapid-repeat** on Detect or Reset USB, despite that being exactly the bug class this PR just fixed.
- **E2E coverage gaps**: no E2E test exercises Save (the real `saveConfig`/`getConfig` round trip), despite Save being one of the four headline features in tasks.md; no E2E coverage of the legacy-resolution fallback path (unit-tested only, with mocked IPC).
- **Remove's "reversible via re-detect" justification for skipping a confirm dialog doesn't hold if the scanner reconnects to a different physical USB port** before re-detect — `upsertScannerRow()`'s fallback matching (bus/device, then port) won't find the old row, creating an orphaned disabled row instead. `BrowseScans.tsx` already uses `window.confirm()` for its destructive action; worth the same convention here, or at least caveating the "reversible" claim in design.md.
- **Raw backend error strings are surfaced verbatim** to a non-programmer lab-tech audience in every failure path (Detect, Reset USB, Remove, Save) — no translation layer to plain-language, actionable copy.
- **Env-status banner text ("Slack wedge alerts: not configured", "libusb recovery: disabled") is jargon with no actionable guidance** for a reader who doesn't already know what a wedge alert or libusb recovery is.
- **Legacy-resolution "touch" gate can permanently disable Save** if the operator's desired value is the fallback-displayed value itself (native `<select>` doesn't fire `onChange` on a no-op reselect) — minor UX trap.
- **Type/shape duplication**: `ScannerRow` and the `handleDetect` scanner-shape both hand-duplicate types already exported elsewhere (`GraviAPI.getScannerStatus()`, `DetectedScanner`), reinforcing the same class of drift that caused Finding 1 above.

## Suggestions

- Named constants for the repeated `1200` fallback-resolution literal and the `3000`ms save-success-message timeout.
- The scanner-status polling `useEffect` recreates its `setInterval` on every tick (because `scanners` is a dependency and every poll produces a new array reference) rather than once on start/stop — works out to the same net cadence by accident, but fragile; consider deriving `anyStarting` as a memo instead.
- Consider splitting `ConfigureScanner.tsx` (437 lines, 3 fairly distinct concerns) into `ScannerTable`/`ResolutionForm`/`EnvBanner` if this page grows in Tier 2.
- Legacy-resolution edge case: auto-mark `resolutionTouched` when the legacy fallback is shown, or add explicit copy ("reselect 1200 to confirm") to avoid the reselect-same-value trap.

## Score / Verdict

Code quality: 7/10. TDD discipline: correctly followed at the right altitude (E2E-first for the real subprocess-timing bug, unit-level for pure logic) — no issue there. The three BLOCKING items above are all realistic operator-facing failure modes on a shared, live rig and should be closed before merge; the IMPORTANT list is a good Tier-2 or immediate-follow-up punch list.

---

Issue #234, opened 2026-05-14
Detect-scanners IPC doesn't spawn scan_workers for newly-discovered rows; user-visible as 'disconnected' until app restart


eberrigan:
## Observed on branch `feature/graviscan-prod`

(Surfaced 2026-05-14 during T6 production-4grid-PCIe test setup, sibling to #230 / #231 / #232.)

## Symptom

When the **Detect Scanners** button discovers a NEW scanner (one whose `usb_port` doesn't already correspond to a `GraviScanner` row in the DB), the IPC creates the new DB row correctly but **does not spawn a `scan_worker` subprocess** for it. The UI then shows the scanner as "disconnected" until the app is closed and re-launched.

Discovering scanners that **already have DB rows** works correctly because their workers were already spawned during the app-startup auto-init.

## Repro (deterministic)

1. Start bloom-graviscan from a state where DB has *N* enabled `GraviScanner` rows.
2. App's `graviscan:auto-init` runs at startup, spawns *N* scan_worker subprocesses (one per enabled row).
3. Physically connect a NEW scanner (or move existing scanners such that some end up on previously-unseen USB paths).
4. Click **Detect Scanners** in the Configure Scanner UI.
5. `graviscan:detect-scanners` finds *N+k* scanners via lsusb.
6. `graviscan:save-scanners-db` upserts: updates the existing N rows; creates k new rows for the new paths.
7. UI re-fetches `getScannerStatus()`, sees *N+k* enabled rows — but the *k* new ones show as **disconnected** (no `scan_worker` running for their `scanner_id`).

Until app restart, the operator can't actually use the newly-discovered scanners. Restarting the app re-runs auto-init, which spawns workers for all enabled rows in DB at that point.

## Root cause

[`src/main/graviscan-handlers.ts`](../blob/feature/graviscan-prod/src/main/graviscan-handlers.ts) `graviscan:save-scanners-db` (~ L436) writes DB rows but does NOT call into the coordinator to spawn workers for newly-created rows. Compare with the app-startup auto-init path (`graviscan:auto-init`, ~ L2189), which loops `db.graviScanner.findMany({where:{enabled:true}})` and spawns one worker per row.

`graviscan:reset-scanners` (~ L1850) kills workers + shuts down the coordinator but doesn't re-spawn them either — so it's not a workaround.

There's no runtime "ensure a worker exists for this scanner_id" check anywhere on the post-discovery path.

[...]

PR #274, opened 2026-07-30
GraviScan: add DB data-layer + granular event model (roadmap Tier 2)

eberrigan:
## Summary

Tier 2 of the [GraviScan renderer roadmap](docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md), backend-only, zero renderer changes. Runs in parallel with Tier 1 (`add-graviscan-configure-scanner-ui`, #273) per the roadmap's dependency graph — the two share one file (`preload.ts`) in disjoint sections, see the roadmap doc's coordination note.

- Ports the missing `database.graviscans.*` / `graviscanSessions.*` / `graviscanPlateAssignments.*` / `graviPlateAccessions.*` IPC handlers (14 total) to `src/main/database-handlers.ts` — the 8 backing Prisma models already existed unchanged, only the IPC layer was missing, blocking roadmap Tiers 4/5.
- Replaces the coordinator's generic `scan-event` bus with granular per-job channels (`scan-started`/`scan-complete`/`scan-error`, each carrying `jobId`/`scannerId`/`plateIndex`), needed by Tier 3's wedge-response UI.
- Audits and resolves issues **#234** (confirmed already fixed on `main`, closing with evidence), **#231** (confirmed not applicable to `main`'s architecture — no per-scanner `grid_mode` column exists), and **#232**'s backend-correctness half (SANE resolution readback, now threaded through TIFF metadata and the `scan-complete` payload).

Every new handler was audited against the specific bug classes the prior GraviScan backend port (PRs #267-272) found — cross-experiment scoping, case-sensitivity, path containment, race conditions — not assumed correct by porting the reference implementation as-is. Two new bugs were found and fixed in the process (not silently reproduced): a missing `experiment_id` scope on `updateGridTimestamps`, and case-sensitive barcode comparison in `checkBarcodeUniqueInWave`.

## Found-and-fixed bug (safety-relevant)

`wiring.ts`'s `setupWedgeDetection()` only ever listened to the generic `scan-event` bus and never saw the coordinator's own directly-emitted `scan-error` events (row-timeout, missing/zero-byte output file) — those failures reached the renderer but never counted toward a wedge's `consecutive_failures` signature. Unifying onto the granular channels closes this gap. This is a real sensitivity change to a safety-relevant path (Slack wedge alerts for time-lapse experiments), flagged explicitly in `design.md` as needing sign-off, with a recommended backtest against the rig's historical failure-log volume before relying on it (tracked as an explicit task, not yet run — needs real rig log access).

## ⚠️ Commit/merge discipline

Commits `84caaa1`→`ca6951b` (event-model migration) **must land together** — `wiring.ts` isn't updated to consume the new channels until the same logical unit as the coordinator's emission change, so any intermediate commit in that range has zero functional wedge detection. Documented inline in `tasks.md` and `design.md`. If squash-merging this PR, this is a non-issue; if merging with history preserved, do not bisect into this range.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npm run test:unit` — 1118 passed, same 7 pre-existing failures as `main`'s baseline (verified by exact test ID/name, not prose description, before and after implementation) — zero regressions. Independently re-verified after implementation, not just self-reported.
- [x] `uv run pytest python/tests/test_scan_worker.py` — 62 passed, 1 pre-existing failure (Windows `fcntl`), 1 skipped — matches baseline
- [ ] **E2E test** (1 retargeted case in `tests/e2e/graviscan-ipc.e2e.ts`: `onScanEvent` → `onScanStarted`) — the retarget is covered by `tsc --noEmit` (imports the real `ElectronAPI` type, which no longer has `onScanEvent`) and mirrors an already-passing pattern elsewhere in the same file, but **could not be executed end-to-end** — requires a full `build:python`+`build:native`+webpack pipeline with no build artifacts available in this session's environment. **Needs a real run in CI or on a machine that can launch Electron before merging.**

## OpenSpec

- Proposal: `openspec/changes/add-graviscan-data-layer-and-events/`
- `npx openspec validate add-graviscan-data-layer-and-events --strict` passes
- `tasks.md`: 101/102 complete — only the E2E-execution checkbox left honestly unchecked (see Test plan above)


eberrigan (commented):

## Review Summary

Strong design/audit discipline (real cross-experiment scoping fixes, case-insensitive barcode matching, genuine transaction-rollback tests, granular per-job event model that closes a real "coordinator-originated errors were invisible to wedge detection" gap) and a thorough unit-test suite. But **CI is currently red on all three platforms** because of a bug in one of this PR's own new E2E tests, confirmed independently by 4 of 5 review lenses — this must be fixed before merge. Several IMPORTANT gaps below are worth closing too, though none are as urgent as the CI fix.

## Blocking Issues

1. **CI is red right now, deterministically, on macOS/Ubuntu/Windows** — a shape-mismatch bug in the new `graviscans.*` E2E block (`tests/e2e/renderer-database-ipc.e2e.ts:2806`, added in commit `188abf0` for task 13.5). The test asserts `expect(timestamps.data).toBe(1)`, but `graviscansUpdateGridTimestamps` returns `{ success: true, data: { updatedCount: result.count } }` — the equivalent *unit* test gets this right (`result.data?.updatedCount`), so this is an isolated E2E-only slip. Task 13.5 explicitly notes these E2E tests were verified only via `tsc`/`lint`/`format:check` and were never actually run before pushing (to satisfy the static IPC-coverage gate, which only checks that a handler is *invoked* somewhere, not that assertions are correct) — this is exactly the failure mode that caution was meant to avoid. Fix: `expect(timestamps.data).toEqual({ updatedCount: 1 })`, then actually run the full E2E suite (not just static checks) before the next push.

## Important Issues

- **`graviPlateAccessionsCreateWithSections` has zero input-type validation**, unlike all 15 sibling handlers (which at minimum guard id fields via `isNonEmptyString`). It's saved from real damage by transaction atomicity and the outer try/catch (a malformed payload throws a plain TypeError, caught and returned as `{success:false}`, not a crash or partial write), but it's the one clear outlier in an otherwise-consistent validation pattern.
- **`electron.d.ts`/`preload.ts` type all 4 new namespaces as `any`**, even though precisely-typed interfaces already exist server-side in `database-handlers.ts` (`GraviScanCreateInput`, `UpdateGridTimestampsInput`, etc.). Every other `DatabaseAPI` entry in this file is precisely typed — there's no precedent for `any` here, and the types just weren't imported into the renderer-facing surface. A caller can pass a malformed payload and TypeScript won't catch it; it'll surface as a generic Prisma error instead.
- **Resolution readback in `scan_worker.py` has no sanity/bounds check.** The `AttributeError` case falls back safely to the requested resolution, but if `x_resolution`/`y_resolution` returns `0`, negative, or another implausible-but-non-throwing value, it flows straight into the TIFF's `IFDRational` tag with only a log warning — no fallback in that case. A misbehaving (not raising) readback silently produces a TIFF with an invalid resolution tag.
- **The wedge-sensitivity backtest / human sign-off called for in `design.md` Open Question 2 and task 11.4a is still outstanding.** Safe to merge into `main` (this migration is strictly additive to detection and the current alerting is Slack-only, no auto-pause/auto-retry), but should gate before this build is deployed to the live rig mid-experiment — folding previously-invisible row-timeouts into the same threshold as subprocess errors risks alert fatigue if they turn out to be frequent and non-wedge-correlated.
- **The scan-complete/scan-error race for the same `(scanner_id, plate_index)` — a path this PR's own fix newly makes reachable** (coordinator-originated errors are no longer invisible) — is handled safely today by `WedgeDetector`'s existing pending-error guard, but is untested by this PR, and it invalidates an explicit "the coordinator surfaces only the FINAL outcome... never both for the same plate" architecture-note comment in `tests/unit/wedge-pipeline-integration.test.ts` that the PR doesn't update.
- **`graviscanPlateAssignmentsUpsertMany` has no test for two concurrent top-level calls** — only single-call atomicity is tested; correctness against a concurrent double-submit depends entirely on implicit SQLite single-writer serialization.
- **Coordinator-originated `scan-error` events lose their `job_id` in the wedge-alert payload** — the 4 direct emit sites only set camelCase `jobId`, never snake_case `job_id`, so `wiring.ts`'s `event.job_id ?? ''` always resolves empty for these. Harmless today (`WedgeDetector` keys state by `scanner_id`, never `job_id`), but a landmine for a future change that does dedupe on it, and it means the Slack alert text is confusingly blank for exactly the failure class this migration was built to surface.
- **Missing cross-experiment/cross-file test coverage** for `graviscanPlateAssignmentsUpsertMany`, `graviPlateAccessionsList`, and `graviscansBrowseByExperiment` — code inspection shows correct scoping in each case, but no test proves isolation holds when two experiments/files coexist, unlike the equivalent tests that do exist for `updateGridTimestamps`/`checkBarcodeUniqueInWave`/`experimentDetail`.
- **PR description states "16 new handlers"; actual count is 14** — minor, but notable given how much this PR's own docs pride themselves on precise, re-verified counts elsewhere.

## Suggestions

- `checkBarcodeUniqueInWave`'s normalization doesn't strip zero-width spaces or apply Unicode normalization (NFC/NFD) — low risk given barcode-scanner input is typically plain ASCII, but worth a note if that assumption ever changes.
- New namespace names (`graviscanSessions`, `graviscanPlateAssignments`, `graviPlateAccessions`) are multi-word camelCase compounds, a minor deviation from the file's existing single-word-namespace convention (action names within each are fine and consistent).
- `scan_worker.py`'s `achieved_resolution` only captures the X-axis readback even though the warning log reports both X/Y — low severity given `_build_tiff_metadata()` already only accepts one scalar and real scanners apply symmetric DPI, but worth a one-line comment documenting the assumption.

## Score / Verdict

Code quality: 6.5/10, docked mainly for shipping red CI on a PR with otherwise strong process discipline. TDD/testing discipline: the unit-test suite is genuinely thorough (real SQLite, deliberate cross-experiment probes, atomicity/rollback tests via a real unique-constraint violation, a well-designed positive/negative pair for the Python readback logic) — but the late-added E2E tests were pushed without ever being executed, and that produced a live, currently-red PR. Fix the one BLOCKING item, get CI green, and this is close to mergeable; the IMPORTANT list is a reasonable immediate-follow-up or Tier-3 punch list.

---

Issue #228, opened 2026-05-09
GraviScan: V600 parallel-scan race + libusb 120s timeout + epkowa stuck-endpoint bug — root cause + 5 candidate fixes


eberrigan:
## TL;DR

We have characterized the failure mode behind GraviScan parallel-scan unreliability. It is a **race condition with a hard 120-second libusb timeout, compounded by a stuck-endpoint bug in epkowa** that requires physical scanner power-cycle to recover.

This issue consolidates 2 days of empirical investigation and proposes 5 candidate fixes (with feasibility analysis), plus a recommended failure-mode UX change.

Cross-references: builds on #125, #171, #177, #182, #185, #226. Related PR: #227 (surfaces the existing libusb-filter shim).

## Failure mechanism (verified)

When N V600 scanners scan in parallel at 1200 DPI, 215×150mm:

1. **All scanimage processes start within microseconds** of each other and issue libusb bulk-IN read requests.
2. **Some shared upstream resource** (PCIe→chipset uplink, scanner power transient on simultaneous lamp ignition, kernel xHCI command queue, or other) means only ONE scanner's bulk read completes inside libusb's 120-second default timeout.
3. **The fastest scanner unit always wins.** In our tests, scanner 1 (a specific physical V600) consistently scanned at 116.7s — just barely under 120s — wherever it was plugged. Scanners 2-5 took >120s under contention and timed out.
4. **The slow scanners hit `LIBUSB_ERROR_TIMEOUT`.** Their bulk-IN endpoint is now in a stalled state.
5. **epkowa's retry logic does NOT call `libusb_clear_halt()`.** SANE debug log: `sanei_usb_open: we already have a bulk-in endpoint (address: 0x81), ignoring the new one`. The endpoint stays halted across all retries within that scanimage process AND across subsequent scanimage invocations on the same scanner.
6. **Only physical scanner power-cycle recovers the device.** USB device reset (`USBDEVFS_RESET`) makes it worse; controller FLR detaches the scanner entirely; reboot works only because it happens after a long enough USB disconnect.

### Evidence — timing data from a single parallel cycle

All 4 scanners started at `19:52:05.988`. Three failed simultaneously at exactly 120.04s:

| Scanner unit | Path | sane_exit | Wall | Outcome |
|---|---|---|---|---|
| **scanner 1** | `17-2` | 19:54:02.689 | **116.7s** | **PASS** |
[...]

PR #277, opened 2026-08-03
GraviScan: add wedge-response UI (roadmap Tier 3)

### LINKED ISSUE #240: GraviScan: in-UI wedge banner when WedgeDetector fires (alongside Slack notification)
Surfaced in PR #237 review (Cluster M — deferred).

## Problem

When the WedgeDetector fires a `wedge-detected` event, the
SlackNotifier posts to the configured webhook (#236). But the UI
shows nothing to the operator watching the screen — there's no
prominent banner, no scanner-row highlight, nothing.

If the operator is actively at the rig and not watching Slack, they
miss the alert until the cycle has finished and they review the log.

## Fix

1. In `src/main/main.ts`, when the `WedgeDetector` emits, forward a
   `graviscan:wedge-detected` IPC event to the renderer in addition
   to the SlackNotifier.notify call.
2. In `src/renderer/GraviScan.tsx` (or a new hook), maintain a list
   of recent wedge events. When non-empty, render a red banner at the
   top of the scanning page:
   - "Scanner X wedged on cycle N (signature: sane_start_invalid).
     Physical AC power-cycle required."
   - Banner is dismissible per-event so an operator can acknowledge
     they've seen it.
3. Optionally: highlight the affected scanner row in the
   ScannerPanel grid with a wedge icon.

## Why deferred

Multi-layer change (main process → IPC → renderer state → UI).
Better as a focused UI PR with consistent treatment for other
operator-paging events (e.g., overtime banner could share the same
component).

## Acceptance

- New `graviscan:wedge-detected` IPC channel + electron.d.ts type
- Renderer listens and accumulates wedge events
- Banner displays scanner_id + signature + power-cycle CTA
- Banner dismissible per-event
- Existing Slack notification flow unchanged

## Related

- PR #237 commit `2edf981` (main.ts wiring of detector + notifier)
- #236 (Slack notification on wedge)

### LINKED ISSUE #244: Wedge response gap: no UI affordance to resume / skip / retry after a wedge alert; irreversible data loss for time-lapse experiments
## Context

During the PR #237 rig validation brainstorm on 2026-05-26, a
`superpowers:brainstorming` subagent investigation confirmed that
the WedgeDetector + SlackNotifier chain (added in PR #237 / #236) has
no follow-on behavior beyond posting the Slack alert. This creates
an operator-facing gap that the new GraviScan user manual currently
has to document as 'permanent data loss, no recovery'.

## What the app does NOT do today

Verified against `fix/v600-wedge-followups@bb97fc3` source by a
parallel subagent investigation:

1. **No pause** — `src/main/main.ts:505-539` `onWedge` callback
   only calls `slackNotifier.notify(enriched)` + a log line. Does
   NOT invoke `scanCoordinator.cancel()`, pause cycle scheduling,
   or signal any subprocess.
2. **No retry** — `_scan_plate`
   ([python/graviscan/scan_worker.py:320](python/graviscan/scan_worker.py))
   swallows the wedge error and continues to the next plate. The
   wedged subprocess STAYS ALIVE in a broken state and re-hits the
   wedge every subsequent cycle.
3. **No UI affordance** — only `graviscan:cancel-scan`
   ([graviscan-handlers.ts:1691](src/main/graviscan-handlers.ts))
   exists. No `graviscan:retry-plate`, `graviscan:skip-plate`,
   or `graviscan:resume` handlers.
4. **No DB record of the wedge** — Slack POST + log line only.
   Detector state is in-memory; cleared on `interval-complete`.

## Why this matters

These are **time-lapse experiments**. Plants grow continuously
during the run. A missed cycle on the wedged scanner cannot be
recovered:

- The plant is in a different state every minute.
- 'Re-running' the missed plate later captures growth from a
  different timepoint, not the missed one.
- For a continuous run, a single un-noticed wedge can wipe out
  hours-to-days of data for the affected scanner's plates.

## Risk analysis (per Subagent D)

- ✅ **Wrong metadata**: safe — `scan-complete` never fires for
  failed scans, so no GraviScan row + no file is created for the
  wedged plates.
- ✅ **Duplicate rows**: safe — only one `scan-complete` per
  (scanner, plate, cycle).
- 🚨 **Missing data (gaps in experiment wave)**: real — the wedged
  scanner produces zero data until manual power-cycle. NO automatic
  re-scan.
- ✅ **Manual SQL cleanup**: not required — the data is just missing,
  nothing to clean up.

## Today's operator workflow

Currently documented in the new GraviScan operator manual as:

- **Option A** — Continue and accept gaps on the wedged scanner.
- **Option B** — Cancel scan, power-cycle the wedged scanner, restart.

Both lose data. The tradeoff math ("if you can power-cycle within
~10 min AND >1 hr experiment remaining, Option B usually loses less
than Option A") leans on operator estimation, not the app helping.

## Proposed UI improvements (any of these would help)

In rough order of value vs. effort:

1. **Per-scanner pause indicator** — when a wedge fires, mark that
   scanner's row in the UI with a 🚨 banner, disable its inclusion
   in subsequent cycles, surface a `Power-Cycled & Resume` button.
   On click: spawn a fresh worker for that scanner_id and re-include
   it in the next cycle.
2. **`wedge_event` DB table** — persist each wedge as a row so
   post-experiment analysis can see exactly which (scanner, plate,
   cycle) tuples were affected. Makes the gap auditable.
3. **Wedge-skip metadata** — for the wedged plate's missed cycles,
   record a placeholder GraviScan row with a `failure_reason=wedge`
   field instead of nothing. Makes the experiment wave
   self-describing.
4. **Auto-cancel-on-N-wedges policy** — operator-configurable. If a
   scanner wedges N times in a session, automatically remove it from
   subsequent cycles + Slack-notify.
5. **Cancel + auto-restart with same parameters** — operator clicks
   one button; app cancels, waits for the operator to confirm
   power-cycle, then resumes from where it left off (no re-entering
   experiment / phenotyper / wave).

## Related

- #228 (V600 wedge investigation — the upstream root-cause)
- #236 (WedgeDetector + SlackNotifier — what this PR built)
- PR #237 (the rig validation walk that surfaced this gap)
- Operator manual spec: `docs/superpowers/specs/2026-05-26-graviscan-user-manual-design.md`
  (Section 11 documents the current workaround)

eberrigan:
## Summary

Auto-pauses a wedged scanner immediately (no operator action required) and adds an app-wide banner with Dismiss / Power-Cycled & Retry actions, closing the "permanent data loss, no recovery" gap for continuous time-lapse GraviScan experiments.

## Changes

- **Auto-pause on wedge detection**: `setupWedgeDetection()`'s `onWedge` callback now calls `coordinator.stopScanner(scanner_id)` immediately, fire-and-forget, not gated behind the Slack notification. A durable `scanLog()` line records the auto-pause alongside the pre-existing wedge-detected line.
- **Forward wedge events to the renderer**: new `graviscan:wedge-detected` push event via an optional `getMainWindow` parameter on `setupWedgeDetection()`.
- **New `graviscan:retry-scanner` IPC handler**: stops (idempotent) and respawns a scanner using a `saneName` rebuilt from a fresh DB read of `usb_bus`/`usb_device` (not a cached value — safe across an intervening `reset-usb`). Checks the scanner's `enabled` flag so a stale banner can't silently revive a scanner the operator explicitly disabled.
- **`buildSaneName()` extracted** from the existing save-scanners-db spawn-on-discovery inline template, shared by both call sites.
- **`WedgeBanner` component** (app-wide, mounted in `Layout.tsx`, gated on `mode === 'graviscan'`): one entry per auto-paused scanner, each with **Dismiss** (no backend call) and **Power-Cycled & Retry** (confirmation-gated — requires an explicit "Confirm Retry" after explanatory text about the power-cycle precondition). A session-scoped indicator shows both total auto-pause events and distinct scanners affected, so a single chronically-re-wedging scanner isn't misread as a fleet-wide problem.
- **`useWedgeEvents` hook**: subscribes to `onWedgeDetected`/`onIntervalComplete`/`onCancelled`, tracks per-scanner entries plus the two session counters.
- In-memory only — no DB schema changes. Deferred: issue #244's `wedge_event` DB table (item 2), wedge-skip placeholder row (item 3), auto-cancel-on-N-wedges (item 4), and session-level cancel+auto-restart (item 5) — all explicitly named as out of scope, not silently dropped.
- **Chore commit included**: fixes pre-existing Prettier formatting drift in `openspec/specs/scanning/spec.md` and `openspec/specs/ui-management-pages/spec.md` (confirmed already broken on `main` itself, unrelated to this feature — whitespace/markdown normalization only) so this PR's Lint job can go green.
- **Test commit included** (added during pre-merge review): real-`ScanCoordinator`-level tests for the exact `stopScanner()` + `addScanner()` sequence `retryScanner()` calls — see Integration Tests below.

Closes #244, #240. Addresses #228's P0 recommendation for the auto-pause half; the resume half is a manual confirmed-retry button rather than #228's fully-automatic USB-re-enumeration design, since this codebase has no hotplug-detection infrastructure (see `design.md` Decision 1 for the full reasoning).

## OpenSpec Change

- Change ID: `add-graviscan-wedge-response-ui`
- Affected capabilities: `scanning` (ADDED — wedge auto-pause, event forwarding, retry-scanner requirements), `ui-management-pages` (ADDED — wedge banner, wedge-response-actions, session auto-pause counter requirements)
- Delta types: ADDED only, no BREAKING changes
- All `tasks.md` items complete: yes (63/63, including task 9.6 added during pre-merge review)
- Proposal went through 4 rounds of 5-agent adversarial review before implementation (see `design.md` for the full history, including a redesign from an earlier manual-"Skip" draft to the current auto-pause design after review found the draft inverted the source issues' intent)

### TypeScript Tests

- [x] Unit tests pass — 1177 passed, 50 skipped (`npm run test:unit`)
- [!] 7 pre-existing failures on `main`, unrelated to this change (verified via file-touch analysis: none of the failing test files — `database-handlers.test.ts`, `config-store.test.ts`, `image-uploader.test.ts`, `scan-coordinator.test.ts` — are touched by this PR beyond adding new, passing tests; the `scan-coordinator.test.ts`/`image-uploader.test.ts` failures are Windows path-separator assertions (`/` vs `\`), `AccessionForm.test.tsx` is a timing-flaky test that passes 10/10 in isolation)
- [x] Unit tests added for new functionality — 93 new/updated tests across `main-wiring.test.ts`, `scanner-handlers.test.ts`, `session-handlers.test.ts`, `register-handlers.test.ts`, `preload-gravi.test.ts`, `scan-coordinator.test.ts`, `useWedgeEvents.test.ts` (new), `WedgeBanner.test.tsx` (new), `Layout.test.tsx`, `App.test.tsx`
- [ ] Coverage threshold check — not run separately; full suite shows no regression

### Python Tests

- [ ] N/A — no Python files touched (this change consumes existing wedge-detector output; it doesn't touch detection logic or hardware-facing code)

### Integration Tests

- [x] `tests/unit/graviscan/scan-coordinator.test.ts` — added two tests exercising the real `ScanCoordinator` class (mocking only `ScannerSubprocess` beneath it, same level as this file's existing coverage) for the exact `stopScanner()` + `addScanner()` sequence `retryScanner()` calls: (1) stop-then-respawn while idle, (2) a retried scanner's `addScanner()` correctly queuing until `cycle-complete` while a *different* scanner is mid-cycle — the real "Retry clicked while the session keeps running" scenario. *(An earlier revision of this PR incorrectly claimed N/A here — see PR comment for the correction.)*
- [ ] No `tests/integration/` (Python-subprocess-level) changes — the new `graviscan:retry-scanner` IPC surface doesn't cross the Node-Python boundary any differently from the existing `addScanner`/`stopScanner` usage that layer already covers

### E2E Tests

- [x] CI's existing E2E suite passes (confirmed green on macOS/ubuntu/windows dev builds)
- [ ] A real wedge scenario (mock hardware to wedge fires to banner to retry) is **not** exercised — `python/graviscan/scan_worker.py`'s mock path has no fault-injection mechanism, and adding one is out of scope for a UI/IPC-wiring change that otherwise touches no Python (documented in `tasks.md` task 9.4). **Tracked for manual rig verification in #279** before this is relied on for an unattended production run.

## Build & Packaging

- [ ] `npm run build:python` — not run; no Python changes
- [ ] Packaged app — not tested; no packaging/build-config/dependency changes

## Linting & Formatting

- [x] Prettier formatting applied (`npm run format:check`) — clean, including the two pre-existing-on-`main` files fixed in a separate chore commit in this PR

## Database Changes

- [x] No database changes

- [x] No platform-specific assumptions introduced (pure TS/React, no new file-path handling)
- [x] CI cross-platform run — green on Linux, macOS, Windows (Python build matrix + integration tests)

## Breaking Changes

- [x] No breaking changes — `setupWedgeDetection()`'s new third parameter is optional with a default; all new IPC channels and preload/`GraviAPI` members are additive

## Related Issues

Closes #244
Closes #240
Related to #228
Follow-up: #279 (manual rig verification of the wedge-response flow)

## Reviewer Notes

- `design.md` in the OpenSpec change directory has the full rationale, including two things worth double-checking: (1) Decision 1's tradeoff analysis of calling `stopScanner()` at wedge-detection time (can occasionally force an in-flight row to fall back to its existing 90s timeout instead of resolving via listeners — accepted as bounded/rare, not a new unbounded risk), and (2) the session-scoped counter's two-number design (event count + distinct-scanner count), added specifically so a single scanner re-wedging after failed retries doesn't read as a multi-scanner incident.
- `tests/unit/graviscan/main-wiring.test.ts` gained a shared `createMockWedgeCoordinator()` factory, refactored across all 13 existing test cases in the `setupWedgeDetection` describe block (previously each built its own bare `EventEmitter` with no `stopScanner` method).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- GraviScan wedge-response UI: auto-pause and operator recovery for USB wedges (#244, #240, #228)
+  - When the `WedgeDetector` fires, the wedged scanner's worker is now automatically stopped and excluded from all subsequent cycles — no operator action required to stop the data loss
+  - App-wide `WedgeBanner` (visible regardless of which screen the operator is on) shows one entry per auto-paused scanner with its signature and error message
+  - Two operator actions per entry: **Dismiss** (acknowledge, no backend effect) and **Power-Cycled & Retry** (confirmation-gated respawn, only after the operator confirms the physical power-cycle is done)
+  - Session-scoped counter shows both total auto-pause events and distinct scanners affected, so a single chronically-re-wedging scanner isn't misread as a fleet-wide problem
+  - New `graviscan:wedge-detected` push event and `graviscan:retry-scanner` IPC handler; durable `scanLog()` entries record every auto-pause and retry attempt
+  - See `openspec/changes/add-graviscan-wedge-response-ui/` for full design rationale
+
 - Idle session timer to prevent scan misattribution in shared lab environments (#102, #116)
   - Main process `IdleTimer` class resets session state after 10 minutes of inactivity
   - Session fields (phenotyper, experiment, wave number, plant age, accession name) cleared on idle
--- a/docs/superpowers/plans/2026-08-03-graviscan-tier3-wedge-response-ui.md
+++ b/docs/superpowers/plans/2026-08-03-graviscan-tier3-wedge-response-ui.md
@@ -35,7 +35,7 @@ sign-off before Tier 3 relies on it, with a recommended backtest first.
 - Zero zero-byte output files among all 28,980 `.tif` files still on disk.
 - Zero individual scan-error log entries of any kind.
 
-Caveat for whoever reads this later: the rig runs the *production* branch, which
+Caveat for whoever reads this later: the rig runs the _production_ branch, which
 doesn't have main's coordinator-level row-timeout/file-verification checks at all
 — so this backtests the underlying physical failure conditions as the best
 available proxy, not an exact replay of the new code path. Small sample (one rig,
--- a/docs/superpowers/plans/2026-08-03-graviscan-tier3-wedge-response-ui.md
+++ b/docs/superpowers/plans/2026-08-03-graviscan-tier3-wedge-response-ui.md
@@ -1,0 +1,57 @@
+# Tier 3 — Wedge-Response UI: Pre-Proposal Notes
+
+**Roadmap:** `docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md`, Tier 3.
+**Depends on:** Tier 2 (merged, PR #274, commit `9805bba`). Does **not** need Tier 1's
+Configure Scanner screen or the full Tier 4 scan screen.
+**Related issues:** #244 ("permanent data loss, no recovery" for wedged time-lapse
+experiments), #240 (no in-UI wedge banner today).
+
+## Why this tier is fast-tracked
+
+Pulled out of the original Tier 4 bundle specifically because #244 describes real,
+irreversible data loss risk on a safety-relevant path, and bundling it into the
+larger, higher-rebuild-risk core scan screen would couple a safety fix's timeline
+to the tier most likely to slip.
+
+## Scope (from the roadmap)
+
+An in-UI wedge banner (#240) with resume/skip/retry affordances (#244), rendered
+against a minimal placeholder scan-state view — merged into the full Tier 4 screen
+once that lands, not held until then.
+
+## Prerequisite decision — resolved 2026-08-03
+
+Tier 2's granular event model made coordinator-originated scan errors (row-timeout,
+missing/zero-byte output file — see `scan-coordinator.ts`'s `scanOnce()`, 4 direct
+`emit('scan-error', ...)` call sites) visible to wedge detection for the first time.
+This was flagged in Tier 2's design doc (Open Question 2) as needing explicit
+sign-off before Tier 3 relies on it, with a recommended backtest first.
+
+**Backtest done, sign-off given.** Pulled the real production rig's logs
+(`graviscan@graviscan-ms-7c56.tail461d0e.ts.net:~/.bloom/logs/`, 2026-05-26 through
+2026-08-01, ~10 weeks):
+
+- Zero wedge alerts have ever fired on this rig in that window.
+- Zero zero-byte output files among all 28,980 `.tif` files still on disk.
+- Zero individual scan-error log entries of any kind.
+
+Caveat for whoever reads this later: the rig runs the *production* branch, which
+doesn't have main's coordinator-level row-timeout/file-verification checks at all
+— so this backtests the underlying physical failure conditions as the best
+available proxy, not an exact replay of the new code path. Small sample (one rig,
+~10 weeks); zero-byte files produced and later cleaned up wouldn't show up in a
+live-disk scan. With that caveat: no evidence of any relevant failure, ever, on
+this rig. **Proceed with Tier 3 without further gating on this question** — don't
+re-litigate it unless new evidence surfaces.
+
+## Next step
+
+Run `/new-feature` (this repo's brainstorm → OpenSpec proposal → `openspec-review`
+5-agent adversarial pass → user approval → `/openspec:apply` TDD → `/pre-merge` →
+PR → `/cleanup-merged` workflow). This worktree is already on its own branch
+(`add-graviscan-wedge-response-ui`, based on `main` @ `2d90a0c`) — a sibling
+`add-wave-scoped-metadata-linking` proposal is running in parallel in a different
+worktree (`../bloom-desktop-wave-metadata-linking`); they touch unrelated files
+(this tier is UI + wedge-detector consumption only, no DB schema), so no
+coordination note is expected, but confirm that's still true once this proposal's
+Impact section is drafted.
--- a/openspec/changes/add-graviscan-wedge-response-ui/design.md
+++ b/openspec/changes/add-graviscan-wedge-response-ui/design.md
@@ -1,0 +1,476 @@
+## Context
+
+No backend or renderer surface for wedge response exists today — confirmed
+by direct code reading, not assumed from the roadmap doc's "No new backend"
+framing (which, read against Tier 1's identical "preload wiring" framing
+for a Tier-1-sized change, turns out to describe forwarding/wiring work,
+not a full data-layer increment — consistent with what this change actually
+does). Specifically:
+
+- `setupWedgeDetection()`'s `onWedge` callback
+  (`src/main/graviscan/wiring.ts:265-277`) only calls `SlackNotifier.notify()`
+  and a log line. `setupCoordinatorEventForwarding()`'s forwarded-event list
+  (`wiring.ts:174-195`) does not include any wedge event, and nothing stops
+  the wedged scanner — per issue #228, the wedged subprocess "stays alive
+  in a broken state and re-hits the wedge every subsequent cycle."
+- `ScanCoordinator`'s public surface
+  (`src/main/graviscan/scan-coordinator.ts`) has `cancelAll()` (session-wide)
+  and `stopScanner()`/`addScanner()` (single-scanner), and nothing scoped to
+  a single job/row/plate.
+- `src/main/preload.ts`'s `graviAPI` object has no wedge-related method or
+  listener.
+- No renderer file references wedge state at all except a static env-config
+  label in `ConfigureScanner.tsx:409` ("Slack wedge alerts:
+  configured/not configured" — unrelated to a live wedge).
+
+Five questions were resolved with the user before/during drafting this
+proposal, since this is a data-loss-prevention safety feature and guessing
+wrong here has real consequences. The first four were resolved before the
+initial draft; the fifth (Decision 1, auto-pause) was resolved after a
+5-agent adversarial review of that draft surfaced a divergence from the
+source issues that the initial clarifying questions had not surfaced.
+
+## Decision 1: Auto-pause on wedge, not manual Skip
+
+**When a wedge is detected, the coordinator automatically stops that
+scanner's worker — no operator click required.** `setupWedgeDetection()`'s
+`onWedge` callback calls `coordinator.stopScanner(evt.scanner_id)`
+immediately (fire-and-forget, not awaited, and not gated behind the
+network-bound Slack notification), in addition to the existing enrich +
+Slack-notify + (new) renderer-forward pipeline.
+
+**Why this superseded an earlier draft.** The first draft of this proposal
+made "Skip" a manual operator action and left a wedged scanner in the
+active rotation (re-failing every cycle) until someone noticed and clicked
+it. Adversarial review checked that draft against the actual source issues
+and found it inverted their intent:
+
+- Issue #228 (the P0 root-cause issue `wedge-detector.ts` itself cites)
+  recommends: "Stop using that scanner automatically (mark as offline;
+  continue with remaining scanners)... Auto-resume when the scanner
+  re-enumerates fresh."
+- Issue #244 item 1 (the item this proposal implements) recommends: "when
+  a wedge fires... disable its inclusion in subsequent cycles, surface a
+  `Power-Cycled & Resume` button. On click: spawn a fresh worker."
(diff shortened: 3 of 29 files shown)


eberrigan:
**Update**: the "Integration Tests: N/A" claim in the PR description was wrong — flagged in review. Every existing test of the `stopScanner()` + `addScanner()` sequence that `retryScanner()` calls (in `main-wiring.test.ts`, `register-handlers.test.ts`, `session-handlers.test.ts`) exercises it against a **fully mocked** `ScanCoordinator`. The actual interaction with the real coordinator — especially `addScanner()`'s mid-cycle queuing, which is exactly the path a Retry click hits while other scanners keep running — was never verified.

Added two tests to `tests/unit/graviscan/scan-coordinator.test.ts` (the one file that already exercises a real `ScanCoordinator`, mocking only `ScannerSubprocess` beneath it):
1. `stopScanner()` removes the worker, `addScanner()` for the same id respawns cleanly while idle
2. `addScanner()` for a retried scanner queues correctly until `cycle-complete` while a **different** scanner is mid-cycle — the actual "Retry clicked while the session keeps running" scenario

Both pass. `tasks.md` updated (task 9.6) to record this addition transparently.

eberrigan (commented):

## Review Summary

Ran a 5-lens adversarial review (code quality/architecture, testing/TDD, scientific rigor/UX, security/cross-platform, behavioral correctness) against the actual diff, cross-checked against `design.md`'s own claims rather than trusting them. Fixed the real bugs the review found in this PR's own new code (3 commits pushed: `6f9e283`, `a816e4e`, `db47d33`). Filed one follow-up issue (#281) for pre-existing coordinator/subprocess-level risks that this PR exercises more often but doesn't introduce and is out of scope to fix here. GitHub Copilot has posted no review comments on this PR.

## Fixed in this round

- **`WedgeBanner.tsx`'s `handleConfirmRetry` had no `catch`** — only `try/finally`. A rejected `retryScanner()` IPC promise (vs. a resolved `{success:false}`) left the operator with zero feedback on a safety-critical action. Added the missing `catch`.
- **Stale-promise race could dismiss the wrong banner entry.** If a scanner re-wedged (superseding its banner entry) while a previous `retryScanner()` call for the same `scanner_id` was still in flight, that stale call's eventual success would delete the *new*, unaddressed entry via `dismiss(scannerId)` (keyed only by scanner_id). Added `dismissIfCurrent()`, which only removes an entry if its `cycle_number`/`timestamp` still match.
- **`retry-scanner` had no concurrency guard.** `addScanner()`/`spawnSingleScanner()` only dedupe concurrent same-scanner calls while a cycle is in flight, not when idle — two concurrent retries for the same scanner could each pass the "already ready?" check and orphan a subprocess. Added an in-flight-retry guard in `session-handlers.ts` scoped to this PR's new call site.
- **`onWedgeDetected`'s preload callback was typed `any`** even though `GraviWedgeEvent` (added by this same PR) already existed for this exact purpose. Retyped.
- **`WedgeBanner` was not actually fixed/sticky**, contrary to design.md Decision 4's "always visible" claim — `Layout.tsx` mounts it inside the scrollable main-content div with no `sticky`/`fixed` CSS, so it could scroll out of view on any page with enough content. Made it `sticky top-0`.
- Disabled Cancel while a retry is in flight (closes a narrow double-click confusion window).
- Added regression tests for all of the above, plus a test for "confirmation resets to unconfirmed when a new wedge supersedes the entry mid-confirmation" — a spec scenario `design.md` claimed was covered but wasn't actually asserted anywhere.
- Corrected `design.md`'s inaccurate claim that testing the `stopScanner()`-mid-row race "would need real subprocess timing" — it wouldn't; the file's existing fake-timer/mocked-EventEmitter infrastructure already proves this reproducible. Documented the real, narrower blocker instead (the shared test helper's `removeAllListeners()` stub).

## Filed as a follow-up (not fixed here) — issue #281

Three real gaps in **pre-existing** `scan-coordinator.ts`/`scanner-subprocess.ts`/`scan_worker.py` code (none of it touched by this PR's diff) that auto-pause/retry now exercise more often:

1. `stopScanner()`'s 5s force-kill can `SIGKILL` the Python worker mid-`image.save()` for a different, healthy plate — no atomic write, no fallback verification for a stripped-listener row.
2. `retryScanner()`'s underlying spawn has no timeout ("SANE open can be slow") — a genuinely hung retry has no UI escape hatch (permanently-disabled "Confirm Retry").
3. `addScanner()`/`spawnSingleScanner()` has no concurrency guard when idle, for *any* caller (this PR's own new call site is now guarded per the fix above, but the underlying coordinator-level race remains for other callers, e.g. a retry racing scanner-discovery's own spawn).

All three require touching production coordinator/subprocess/Python code, which is out of scope for a tier explicitly scoped to "wedge-detector consumption + UI, no coordinator changes" — filed to track rather than silently drop, matching this repo's existing convention (#279).

## Flagging for your judgment call, not unilaterally changed

**`retryScanner()` resolves `{success: true}` even when the respawned worker silently fails to come online** (spawn failures surface via a separate `scanner-init-status` event this tier doesn't correlate back to the retry attempt). This is already an explicit, named "accepted limitation" in `design.md` from the proposal's own adversarial review round — but two independent reviewers in *this* round (security, scientific rigor) both re-flagged it as more severe than "accepted": an operator can believe a scanner is fixed and back in rotation for the rest of a multi-day run when it silently isn't, which is close to the exact failure mode (#244) this whole feature exists to prevent. I didn't reverse an already-deliberated scope decision unilaterally — flagging it here since two fresh, independent lenses converged on it being underweighted. Worth deciding whether this becomes a fast-follow (correlating `scanner-init-status` before declaring retry success) rather than staying deferred indefinitely.

## Suggestions not acted on (low priority, noted for awareness)

- Session counter copy ("3 auto-pause events across 1 scanner this session") doesn't state the total active-scanner count, so severity can't be fully calibrated from the copy alone.
- No audio/OS-level alert on the banner — acceptable given issue #240's narrower "operator on a different in-app screen" scope, Slack remains the out-of-band channel for genuinely unattended operation.
- A stale banner isn't auto-cleared by an out-of-band `reset-usb()`, so a subsequent stale Retry click causes a redundant (but harmless/idempotent) respawn cycle.
- No commit-level evidence of granular per-section TDD (all 9 implementation sections landed in one commit) — the test *content* itself reads as authored test-first, but the process claim can't be verified from `git log` alone.

## CI / merge status

All CI jobs were green as of this review round except `Test - E2E Dev Build (ubuntu-latest)`, which was still finishing — recommend re-checking `gh pr checks 277` before merge. New commits since the last CI run will need their own green run.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), triaged and independently verified against source before any fix was applied.*

Issue #249, opened 2026-06-09
Multiple bloom-desktop instances can run concurrently (affects both GraviScan and Cylinder workflows)


eberrigan:
## Problem

The bloom-desktop Electron app can be launched multiple times, producing two (or more) concurrent main processes. Both processes initialize their own subprocesses (Python, ScanCoordinator, CameraProcess, DAQProcess) and both open the same SQLite database at `~/.bloom/data/bloom.db`.

This affects **both the GraviScan and Cylinder workflows** — they share the same Electron `main.ts` and main process. There is no subsystem-level isolation: every singleton subprocess (PythonProcess, CameraProcess, DAQProcess, ScanCoordinator) gets duplicated when a second instance launches.

## Repro

1. Launch bloom-desktop
2. Launch bloom-desktop a second time
3. Observe: a second window appears, owned by a separate main process. Both windows operate independently against the same `bloom.db`.

## Why it matters

Concurrent instances cause:

- **Database race conditions / corruption risk** — both processes open SQLite at `~/.bloom/data/bloom.db` and perform concurrent reads/writes via Prisma. WAL mode mitigates but doesn't eliminate the risk.
- **GraviScan subprocess conflicts** — each main process spawns its own ScanCoordinator → scanner-worker subprocesses. Two coordinators competing for the same USB devices (epkowa SANE backend, which already has issues with concurrent device claims — see #228) will produce unpredictable scan behavior.
- **Cylinder workflow conflicts** — duplicate PythonProcess / CameraProcess / DAQProcess instances attempting to control the same hardware.
- **Settings/state divergence** — config-store and session-store reads from one process won't see writes from the other.

## Root cause

`src/main/main.ts` does not call `app.requestSingleInstanceLock()`. Verified by:

```bash
grep -rn "requestSingleInstanceLock\|second-instance\|gotTheLock" src/
# returns nothing
```

This is the standard Electron API for ensuring single-instance behavior. Without it, every `bloom-graviscan` invocation creates a new top-level process.

## Proposed fix

Add the standard Electron single-instance pattern to `src/main/main.ts` after the `app` import and before any subprocess initialization or `app.whenReady()` call:

```typescript
const gotTheLock = app.requestSingleInstanceLock();
if (!gotTheLock) {
  app.quit();
} else {
  app.on('second-instance', () => {
    // Someone tried to launch a second instance — focus the existing window
    const win = BrowserWindow.getAllWindows()[0];
[...]

Issue #198, opened 2026-04-23
Python Backend Status panel hardcoded for CylinderScan (shows camera/DAQ warnings in GraviScan mode)


eberrigan:
## Problem

The \"Python Backend Status\" panel on the Home page (`src/renderer/components/PythonStatus.tsx` via `checkHardware` IPC) always reports on CylinderScan hardware — camera (Basler Pylon) and DAQ (NI-DAQmx) — regardless of which scanner mode is active.

In GraviScan mode this produces scary-looking warnings:

```
Camera: [WARN] Library installed, no devices found
DAQ:    [WARN] Library installed, no devices found
```

These warnings are irrelevant — GraviScan uses SANE/TWAIN flatbed scanners via the Epson backend, not Pylon/DAQ. A user sees two amber warnings on their Home page and reasonably assumes something is broken.

Meanwhile, there is no equivalent check for the GraviScan backend: whether SANE libraries are installed, whether the Python graviscan subprocess can launch, whether Epson scanner drivers are available. A real issue there (e.g., missing `sane` Python package, broken epkowa backend) would never surface in the Home status panel.

## Root cause

- `python/ipc_handler.py::check_hardware` only checks `pypylon` (camera) and `nidaqmx` (DAQ).
- No Python-side `check_graviscan_hardware` equivalent exists for SANE / Epson devices.
- The renderer shows the hardware status unconditionally — no mode-aware filtering.

## Proposed fix

1. **Add a `check_graviscan_hardware` path in Python** that reports:
   - `sane` Python package installed / not installed
   - `graviscan.scan_worker` importable
   - SANE backend available (`sane.init()` or equivalent)
   - Mock mode flag echoed back

2. **Route via `scanner_mode` config** — `ipc_handler.check_hardware` should read the configured mode (or accept it as a param) and return the appropriate hardware report:
   - CylinderScan → camera + DAQ (current behavior)
   - GraviScan → SANE + Epson subprocess
   - Both checks skipped when mock flag is on, with a visible \"Mock mode\" indicator instead

3. **Renderer: show only mode-relevant warnings** — `PythonStatus.tsx` reads `mode` from `useAppMode` and labels the panel accordingly (\"Flatbed Scanner\" for graviscan, \"Camera / DAQ\" for cylinderscan).

## Acceptance criteria

- [ ] GraviScan mode Home page shows \"SANE / Epson scanner\" hardware status, not \"Camera / DAQ\"
[...]

Issue #97, opened 2026-03-02
Replace any types in ImageUploader with proper types from bloom-fs/bloom-js


eberrigan:
## Summary

`src/main/image-uploader.ts` uses `any` type annotations for 4 dynamically-imported Supabase/bloom dependencies, with `eslint-disable` comments to suppress warnings. These should be replaced with proper types from the packages.

## Current Code

```typescript
// src/main/image-uploader.ts lines 95-102
// eslint-disable-next-line @typescript-eslint/no-explicit-any
private supabase: any = null;
// eslint-disable-next-line @typescript-eslint/no-explicit-any
private uploader: any = null;
// eslint-disable-next-line @typescript-eslint/no-explicit-any
private store: any = null;
// eslint-disable-next-line @typescript-eslint/no-explicit-any
private uploadImagesFn: any = null;
```

## Expected

Use proper types:

```typescript
import type { SupabaseClient } from '@supabase/supabase-js';
import type { SupabaseUploader, SupabaseStore } from '@salk-hpi/bloom-js';
import type { uploadImages } from '@salk-hpi/bloom-fs';

private supabase: SupabaseClient | null = null;
private uploader: SupabaseUploader | null = null;
private store: SupabaseStore | null = null;
private uploadImagesFn: typeof uploadImages | null = null;
```

**Note:** Type-only imports (`import type`) do not cause the runtime loading issue that motivates the dynamic `await import()` pattern. Only runtime imports need to be dynamic.

## Complexity

- Verify that `@salk-hpi/bloom-js` and `@salk-hpi/bloom-fs` export these types
- If types are not exported, may need to use `Awaited<ReturnType<typeof import(...)>>` pattern
- Already partially tracked in `openspec/changes/fix-upload-database-registration/tasks.md` Phase 3.1

## Context

Found during PR #92 code review. Deferred because it requires investigating bloom-fs/bloom-js type exports.

eberrigan:
Fixed in https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/pull/280 (roadmap Tier 1, merge commit `995c1df`; re-confirmed unchanged by Tier 2, PR #304). Verified on `main`: no `: any` types remain in `image-uploader.ts`. Closing — confirmed by reading current code during a 2026-08-06 roadmap audit.

Issue #96, opened 2026-03-02
Preload listener cleanup: 8 of 12 on* methods lack cleanup functions


eberrigan:
## Summary

In `src/main/preload.ts`, 8 of 12 `on*` event listener methods do not return cleanup functions. This means renderer components cannot properly unsubscribe from IPC events, leading to potential memory leaks if components unmount and remount.

## Current State

**Methods WITH cleanup (4/12):**
- `camera.onFrame` (line 88-91) — returns `() => ipcRenderer.removeListener(...)`
- `scanner.onProgress` (line 143-148)
- `scanner.onComplete` (line 150-155)
- `scanner.onError` (line 157-161)

**Methods WITHOUT cleanup (8/12):**
- `python.onStatus` (line 53-56)
- `python.onError` (line 58-60)
- `camera.onTrigger` (line 76-78)
- `camera.onImageCaptured` (line 79-83)
- `daq.onInitialized` (line 108-110)
- `daq.onPositionChanged` (line 111-116)
- `daq.onHome` (line 117-119)
- `daq.onError` (line 120-122)

## Expected Behavior

All `on*` methods should return a cleanup function that removes the listener, matching the pattern used by `camera.onFrame` and the `scanner.on*` methods:

```typescript
onStatus: (callback: (status: string) => void) => {
  const listener = (_event: unknown, status: string) => callback(status);
  ipcRenderer.on('python:status', listener);
  return () => ipcRenderer.removeListener('python:status', listener);
},
```

## Impact

- **Memory leaks**: Components that subscribe to events without cleanup accumulate stale listeners
- **Type safety**: The `PythonAPI`, `CameraAPI`, `DAQAPI` interfaces in `src/types/electron.d.ts` should be updated to return cleanup functions
- **Consistency**: Current code has two patterns for the same concept

## Files to Modify

- `src/main/preload.ts` — Add cleanup return values
- `src/types/electron.d.ts` — Update return types from `void` to `() => void`
- Renderer components using these APIs — Use cleanup in `useEffect` returns

## Context

Found during PR #92 code review. Deferred because the fix touches 8 methods + types + renderer components.

eberrigan:
[...]

PR #280, opened 2026-08-03
CylinderScan: correctness & security hardening (finalization roadmap Tier 1)

eberrigan:
## Summary

Tier 1 of the [CylinderScan finalization roadmap](docs/superpowers/plans/2026-08-03-cylinderscan-finalization-roadmap.md) (5-tier plan to close feature-parity gaps against the retired `bloom-desktop-pilot` app, fix known bugs, and get CylinderScan ready for Windows deployment). This PR is the correctness-and-security hardening tier — de-risking the codebase before Tier 2 touches upload/delete data-integrity work:

- **#249** — `app.requestSingleInstanceLock()`, so two instances can't run concurrently against the same SQLite database and hardware (a real data-corruption risk on lab machines). Extracted into testable `src/main/single-instance.ts` helpers.
- **#96 + #198** (combined — both touch `PythonStatus.tsx`'s same `useEffect`) — the 8 preload listeners missing cleanup functions now have them; `PythonStatus.tsx`'s real leak (its effect never consumed the cleanup) is fixed, and the component is now mode-aware (gates its *effect body*, not just render output, per React's Rules of Hooks — a render-only gate would still fire the IPC calls in graviscan mode).
- **#47** — `PythonProcess.sendCommand()` now correlates requests to responses by id via a `pendingRequests` map, instead of a single-shot `once('data'/'error', ...)` pair that could resolve/reject the wrong in-flight call. Found and fixed a real concurrency bug during review: the camera-streaming background thread was calling `send_error()` in a way that could misattribute its own async errors to whatever unrelated command happened to be in flight on the main thread.
- **#93** — Replaces `webSecurity: false` with a custom `bloom-scan://` privileged protocol, serving scan images with explicit, boundary-aware path-containment validation (`src/main/scan-protocol.ts`) instead of disabling the renderer's web security entirely. Found and fixed a real containment bug during review (a naive prefix check would accept a sibling directory sharing a name prefix) and a real URL round-trip bug during implementation (the Windows drive-letter URL convention needs explicit handling on the way back to a native path).
- **#97** — The 4 `any`-typed `ImageUploader` fields now use their real, already-published types from `@salk-hpi/bloom-js`/`@salk-hpi/bloom-fs`. No behavior change.
- **#40** — Investigation confirmed the scanner-state race this issue describes isn't reachable under the current single-threaded architecture. Per explicit user decision, this is closed as verified-not-reproducible (a regression test + a defensive runtime check that survives `-O`), not fixed with a `threading.Lock`.

This also bundles the roadmap doc, since this branch is where the roadmap was drafted before diverging into Tier 1's implementation.

The proposal (`openspec/changes/harden-cylinderscan-tier1/`) went through **4 rounds** of a 5-agent adversarial `openspec-review` pass before implementation started. Rounds 1-3 found and fixed real issues (a genuine concurrency bug, a false premise about a nonexistent "PR #41 delay," a React Rules-of-Hooks bug, a path-containment prefix-collision bug, a bare Python `assert` that `-O` would strip, a cross-platform `path.win32`/`path.posix` bug, and several existing tests that would otherwise break silently); round 4 converged clean across all 5 lenses. See `design.md`'s Decision sections for full detail on each.

## Notable implementation findings (see `design.md` for full detail)

- **Streaming-thread error misattribution** (#47): `streaming_worker()` runs on a genuine background thread and calls `send_error()` — without a fix, an async streaming error could reject an unrelated in-flight command. `send_error()` now takes a `tag_request` flag; the streaming thread passes `False`.
- **Path-containment prefix collision** (#93): a naive `startsWith(scansDir)` check would wrongly accept `bloom-scans-archive` as contained within `bloom-scans`. Fixed to a boundary-aware check.
- **URL round-trip bug** (#93, found during implementation, not review): `pathToFileUrl()` prepends a leading slash before a Windows drive letter (the standard `file://` convention) — the protocol handler has to strip it back off before calling `path.resolve()`, or `path.resolve()` treats `C:` as a literal path segment under root instead of a drive letter. Caught by the new `scan-protocol.test.ts`'s freshness test failing unexpectedly.
- **Cross-platform test determinism** (#93): `resolveScanPath()` takes an injectable `path` module (defaulting to the ambient one) so Windows-specific behavior is testable deterministically on Linux/macOS CI, not just when actually running on Windows.
- **No real PR #41 delay** (#47): an earlier draft of this proposal claimed to remove a "hardcoded ~1-second delay" per issue #47's own text — that claim doesn't match what PR #41 actually shipped (an Apply-Settings UI gate, not a delay). Confirmed via `gh pr view 41`; no removal action taken, just a note so no one goes looking for something that isn't there.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean (2 pre-existing warnings on untouched files, confirmed via empty diff against `main`)
- [x] `npm run test:unit` — 1173 passed, 50 skipped, same 7 pre-existing failures across 5 files as `main`'s baseline (confirmed identical before any Tier 1 work began and throughout) — zero regressions
- [x] `uv run pytest python/tests/` — 273 passed, 6 skipped, same 6 pre-existing Windows-temp-file-lock failures as baseline — zero regressions
- [x] `ruff check` / `black --check` / `mypy` on all changed Python files — clean
- [ ] **E2E tests** (`scan-preview.e2e.ts`, `scan-directory-format.e2e.ts` updated for the new `bloom-scan://` scheme) — written, lint/tsc-clean, but could not be executed in this sandboxed development environment (consistent with prior sessions' note about this environment). **Needs a real run in CI before merging.**

## OpenSpec

- Proposal: `openspec/changes/harden-cylinderscan-tier1/`
- `npx openspec validate harden-cylinderscan-tier1 --strict` passes
- `tasks.md`: 34/34 tasks complete


eberrigan (commented):

## Review Summary

5-lens adversarial review (code quality/architecture, testing/TDD, scientific rigor/UX, security/cross-platform, behavioural correctness/edge cases) of the current diff, including the post-open `bloom-scan://` URL fix and the merge with `main`. Each lens read the actual checked-out source, not just the diff. Three independent, cross-verified BLOCKING issues survived a second manual verification pass (I re-checked each one directly against the source before including it here) — this PR is not merge-ready as-is despite CI being fully green. Green CI here proves the fix for the *original* reported bug works; it does not cover the newly-found issues below, none of which are exercised by existing tests.

## Blocking Issues

1. **`focusExistingWindow()` (`src/main/single-instance.ts:15-21`, new in this PR's #249 work) has no `isDestroyed()` guard, unlike every other one of the 13+ `mainWindow` call sites in `main.ts`.** On macOS, `window-all-closed` deliberately keeps the app alive without resetting `mainWindow` to `null` (`main.ts:1281-1285`) — closing the window then triggering a second launch fires `second-instance` → `focusExistingWindow(mainWindow)` against a destroyed `BrowserWindow`, and `.isMinimized()`/`.focus()` throw uncaught. `tests/unit/single-instance-lock.test.ts` covers `null`/minimized/live cases but not destroyed. Fix: `if (!win || win.isDestroyed()) return;`.

2. **`PythonProcess.restart()` (`src/main/python-process.ts:194-198`) has no re-entrancy guard, and this PR's own new `rejectAllPending()`-on-`exit` wiring turns that into a real cross-generation race.** `restart()` fires `kill()` and moves on without waiting for the old process's actual OS exit. If that stale `exit` event lands *after* a new process has started and a new `sendCommand()` is already in flight, `rejectAllPending()` rejects the new, legitimately-pending request too — misattribution of exactly the kind #47 was built to eliminate, now reintroduced at the restart boundary. Compounding factor: the "Restart Python" button (`PythonStatus.tsx:89-100,190-195`) has no `disabled`/in-flight guard, so a user double-click can trigger this directly. Confirmed via `git merge-base --is-ancestor` that this machinery (`pendingRequests`, `rejectAllPending`, the `exit` wiring) is new in this PR, not pre-existing. Zero test coverage for `restart()` in any form.

3. **`openspec/changes/harden-cylinderscan-tier1/specs/scan-preview/spec.md` (and `tasks.md` 4.3/4.4) still document the pre-fix, Chromium-corrupted URL format** (`bloom-scan:///C:/Users/foo/bar.png`) as the required behavior, even though the actual shipped fix uses `bloom-scan://local-file/?path=<encoded>`. This spec-delta was never touched by the fix commit. On archive, this becomes the permanent, durable spec of record — a future engineer implementing against it could reintroduce the exact bug this PR just fixed, with the project's own source of truth pointing them at the wrong answer. `openspec validate --strict` only checks structural formatting, so it can't and didn't catch this content drift.

## Important Issues

- No durable/enforced rule (only prose in `design.md`, which isn't part of the archived spec) requiring a real Electron E2E run before merging future changes to `scan-protocol.ts`/`file-url.ts` — the exact gap that let the original bug through 4 rounds of spec review.
- `src.toContain('bloom-scan://')`/`.toContain('1.png')` assertions in the E2E suite are decorative for this bug class (they check the raw JS-constructed string, not the actual resolved fetch) — the real regression guard is the `imageNotFound`/`imgElement` count checks. Worth a comment so this isn't misread later.
- PR description's test-plan checklist still says E2E "could not be executed... needs a real run in CI" — stale; it's been run locally and is green in CI now.
- `python/ipc_handler.py`'s `handle_command` clears `_current_request_id` in a `finally` *before* the outer `except Exception` in `run_ipc_loop` reports it — untagged top-level errors reject *every* pending request instead of the one that actually failed, for the `ping`/`get_version`/`check_hardware`/unknown-command paths (sub-handlers already swallow their own exceptions, so this is narrower than it sounds, but still a real gap in #47's precision).
- No main-process diagnostic logging on the 403/404 paths in `scan-protocol.ts` — this is not hypothetical, it's what made this week's real "Image not found" misconfiguration take significant time to root-cause. A `console.warn` with the resolved path + configured `scansDir` + rejection reason is a cheap fast-follow.
- Symlink escape isn't covered by `resolveScanPath()`'s lexical containment check (3 lenses flagged this independently). `design.md:17` explicitly accepts this as an out-of-scope tradeoff given the current single-user-machine threat model, which is a reasonable call — but GraviScan's sibling `read-scan-image` handler already does realpath-based containment, so there's an inconsistency between the two image-serving code paths worth reconciling in a later tier.
- `scans_dir === ''` (e.g. from a corrupted `.env`) fails open to `process.cwd()` as the containment root rather than failing closed — cheap fix (`if (!scansDir) return null;`).
- A `second-instance` event racing very early startup (`mainWindow` still `null`) drops the focus request with no retry once the window exists — narrow, but undermines #249's whole "second launch feels safe" premise.
- No true two-process integration/E2E test exists for #249 — only the pure helper functions (`shouldQuitAsSecondInstance`/`focusExistingWindow`) are unit-tested; nothing spawns a real second Electron process, despite the three target platforms implementing single-instance locking via genuinely different OS mechanisms.
- A pre-existing bare `assert` in `streaming_worker()` (`ipc_handler.py:445`) sits right next to lines this PR touched, contradicting the bare-assert-under-`-O` fix pattern this same PR establishes elsewhere in the file.

## Suggestions

- Raw IPC `id` field (from #47) flows through `any`-typed responses into `ScannerProcess`'s `scanner:complete` emission — currently harmless (verified `saveScanToDatabase`/metadata-writing use explicit field selection, not spread), but worth a guard or comment so a future refactor doesn't accidentally leak/rely on it.
- `pathToFileUrl()`'s `encodeURIComponent` call is uncaught; a malformed unpaired-surrogate string (unlikely from a real path, possible from corrupted DB data) would throw synchronously during React render with no error boundary in `ScanPreview.tsx`.
- The `local-file` host in `bloom-scan://` URLs is purely cosmetic (never validated) — the JSDoc phrasing could be read as implying it's part of the security contract; worth clarifying it isn't.
- Missing `path` query param falls back to `''` → resolves to cwd → (almost always) 403 for the wrong reason. An explicit 400 would be clearer.
- macOS's default case-insensitive-but-case-preserving filesystem isn't handled by `resolveScanPath()`'s Windows-only case-fold branch — same category of behavior as Windows per the design doc's own stated rationale, just unaddressed on macOS. Low risk today since paths are app-generated on both sides.
- `getScansDir()` does synchronous blocking disk I/O (`fs.readFileSync`) on every single image request on the main process's one thread — fine at current scale, worth a TTL cache if thumbnail grids grow.
- Consider `X-Content-Type-Options: nosniff` on the protocol handler's responses as cheap defense-in-depth, since Content-Type is derived from extension only, never sniffed from content.

## What held up well

- The core #93 fix (query-param URL encoding) is correct and well-tested against real `URL`/`URLSearchParams` parsing, not just hand-rolled string checks — verified it round-trips spaces, `#`, `&`, `%`, `+`, and non-ASCII characters correctly, and that `..`-based traversal, double-encoding, and null-byte tricks don't bypass the containment check.
- `#47`'s streaming-thread misattribution fix (`tag_request=False`) is correctly designed to avoid the race entirely rather than lock around it.
- `#198`'s mode-gating in `PythonStatus.tsx` correctly follows React's Rules of Hooks (effect body gated, not just render), with correct cleanup-before-next-effect-run semantics on a mode change.
- All 8 previously-broken preload listeners now correctly return cleanup functions (4 of them — `camera.onTrigger/onImageCaptured`, `daq.onPositionChanged/onHome` — have zero current renderer consumers, so this fix is currently prophylactic rather than closing an active leak, which is fine).
- No exploitable path-traversal, injection, or IPC-spoofing vector was found from the security lens specifically — the flagged items are concurrency/completeness gaps, not access-control bypasses.

---

eberrigan (commented):
> **Verdict: REQUEST_CHANGES → addressed** (posted as comment — cannot request changes on own PR). Round 2 convergence check on commit `9a78f43` (the round-1 fix commit). All 3 round-1 blocking fixes independently confirmed correct by every lens, including hand-traced mutation testing. Found 2 more small, real, in-scope issues (now fixed in `0e5df6d`) and one much larger pre-existing gap explicitly judged out of scope and filed as a follow-up issue.

## Review Summary

5-lens re-review of PR #280's diff, explicitly instructed to re-verify round 1's 3 fixes rather than go easier on a follow-up round, and to do a genuinely fresh adversarial pass. All 3 round-1 fixes (`focusExistingWindow()`'s `isDestroyed()` guard, `PythonProcess`'s generation-tagging + restart memoization, the corrected OpenSpec spec-delta) were unanimously confirmed correct — one reviewer went as far as hand-tracing specific mutations of the fix (removing just the generation check, removing just the memoization, inverting the comparison) to confirm the new tests actually discriminate the real fix from plausible-but-wrong alternatives. No reviewer found a way to break any of the three.

Round 2 surfaced 2 additional real, narrowly-scoped bugs (both now fixed) and one much larger, genuinely serious, but pre-existing gap outside this PR's diff (filed as a follow-up issue rather than expanding this tier's scope).

## Confirmed Correct (Round 1 Fixes)

- `focusExistingWindow()`'s `isDestroyed()` guard — confirmed consistent with all 16 `mainWindow` call sites in `main.ts` (not just the 13 originally cited).
- `PythonProcess`'s generation-tagging + `restart()` promise memoization — hand-traced against 4 specific edge cases (direct `stop()`+`start()` without `restart()`; back-to-back `restart()` calls after the promise clears; `start()` throwing synchronously inside `performRestart()`; a `restart()` call when the process was never started) — all handled correctly.
- The `scan-preview` spec-delta now matches the shipped query-param URL format byte-for-byte against the actual source, with the "why not triple-slash" rationale preserved.

## Fixed This Round

1. **`restartPython()` (`PythonStatus.tsx`) silently reported success on a failed restart.** `window.electron.python.restart()` never rejects — it always resolves, reporting failure via `{ success: false, error }`. The component ignored this and unconditionally showed "Restarted." Fixed to check `result.success` and surface the real error. Also corrected `electron.d.ts`'s `restart()` return type, which was missing the `error` field the real handler has always returned.
2. **`stop()` didn't proactively reject pending requests — an unintended side effect of round 1's generation guard.** Before round 1, any exit event unconditionally rejected all pending requests (the misattribution bug, but incidentally also failing genuinely-orphaned requests fast). Round 1's fix correctly stops a *stale* exit event from touching a *new* generation — but as a side effect, a truly-orphaned old-generation request (pending at the moment `stop()` was called) was no longer proactively rejected, and would silently wait out the full 3-minute timeout instead. `stop()` now rejects pending requests itself, at the one point where we know for certain no further response is coming.

Both fixed via TDD (failing test first, confirmed red for the right reason, then implementation). Also added a regression test confirming generation-guard protection isn't `restart()`-specific — traced by hand and found already correct, just untested.

## Filed As a Follow-Up, Not Fixed Here

**A Python subprocess crash mid-scan strands the UI in "Scanning..." forever** ([#287](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/287)). `ScannerProcess.scan()` doesn't wrap its `sendCommand()` call in try/catch, so a process death mid-scan never fires `'complete'`/`'error'`; `CaptureScan.tsx` never checks `scanner.scan()`'s resolved `{success, error}`, so `isScanning` never returns to `false` and the operator sees no error, with a stray `metadata.json` left on disk describing a scan that never happened. This is real and serious, but it predates this PR, spans files entirely outside this diff (`CaptureScan.tsx`, `scanner-process.ts`), and isn't caused by the restart-race work this review was checking — the new, hardened restart button just makes an existing gap one click easier to reach. Consistent with this proposal's existing precedent (e.g. deferring #198's full GraviScan hardware-status equivalent), this is scoped as its own follow-up rather than folded into this hardening tier.

## Remaining Suggestions (non-blocking)

- No test yet for the restart-rejects-then-button-reenables path specifically (distinct from the restart-succeeds path already tested).
- `python/ipc_handler.py`'s exception path can lose request-id attribution for a bug in the top-level `ping`/`get_version`/`check_hardware` dispatch itself (narrow; degrades to a safe "reject everyone" rather than a misattribution).
- Minor: 6 of the 8 preload listeners fixed for cleanup (#96) have no current renderer consumers — the fix is currently prophylactic, which is fine.

---
*Round 2 review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness)*

Issue #281, opened 2026-08-03
GraviScan: pre-existing coordinator/subprocess safety gaps exposed more often by auto-pause/retry (SIGKILL corruption, no-timeout hang, spawn race)


eberrigan:
## Context

PR #277 (`add-graviscan-wedge-response-ui`, roadmap Tier 3) made
`ScanCoordinator.stopScanner()` and `addScanner()` fire automatically and
more frequently than before: `stopScanner()` now runs on every wedge
detection (previously never called automatically), and `addScanner()` now
runs from a new operator-facing `retry-scanner` IPC handler (previously
only from scanner discovery). Neither method's own implementation changed
— but the PR's adversarial review round surfaced three pre-existing gaps
in `scan-coordinator.ts`/`scanner-subprocess.ts`/`scan_worker.py` that this
new, more-frequent exposure makes materially more likely to matter. None
of these are introduced by PR #277's diff; all three predate it.

## 1. `stopScanner()`'s 5s force-kill can SIGKILL the Python worker mid-write, corrupting a healthy plate's output file

`stopScanner()` calls `sub.shutdown(5000)`
(`scan-coordinator.ts`), which sends `quit()` and force-kills via
`SIGKILL` after a fixed 5000ms if the process hasn't exited
(`scanner-subprocess.ts`). But the Python worker's command loop only
checks for `quit` between whole scan-command batches, not between
individual plates within a batch, and not at all during an in-progress
plate scan (`scan_worker.py`'s `run()`/`_handle_scan()`). `image.save()`
writes directly to the final output path with no temp-file/atomic-rename.

A wedge signature firing on one scanner does not mean that scanner's
*current* plate write is done — if `stopScanner()`'s 5s grace period is
exceeded (routine V600 scans commonly take longer than 5s), the resulting
SIGKILL can truncate an in-progress `image.save()` write for a plate that
was otherwise fine, leaving a non-empty, invalid TIFF at the exact
filename a downstream pipeline expects for that timepoint. This corrupt
file is invisible to existing safety nets: `stopScanner()`'s
`removeAllListeners()` strips the row's completion listeners, so
`scanOnce()`'s per-row file-verification loop skips existence/size checks
entirely for a `null` result (see `scan-coordinator.ts`'s `if (!result)
continue;`).

**Suggested directions (not scoped/prioritized here):** write-then-atomic-rename
in the Python worker (`os.rename()` into the final path only after a
[...]

Issue #279, opened 2026-08-03
GraviScan: manual rig verification needed for wedge-response UI (auto-pause + retry)


eberrigan:
## Context

PR #277 (`add-graviscan-wedge-response-ui`, roadmap Tier 3) adds auto-pause
and an operator-facing Retry action for V600 USB wedges, closing #244/#240.
The change is covered by 91 unit tests, but — as documented in the
OpenSpec change's `tasks.md` (task 9.4) — a real end-to-end wedge scenario
cannot be exercised via mock hardware or CI's E2E suite:
`python/graviscan/scan_worker.py`'s mock path has no fault-injection
mechanism, so nothing in CI ever actually triggers `sane_start: Invalid
argument`, a 120s-zero-byte I/O stall, or consecutive failures on a real
device.

## What the automated tests DON'T cover

- Whether a real wedge (on real V600 hardware) actually triggers
  `coordinator.stopScanner()` and stops the physical scanner mid-session
  (unit tests only exercise this against a mocked `ScanCoordinator`)
- Whether the `WedgeBanner` renders correctly and stays responsive while a
  real multi-scanner interval session is running
- Whether "Power-Cycled & Retry" actually respawns a real SANE subprocess
  against the correct USB port after a genuine physical power-cycle, and
  whether the respawned worker successfully resumes scanning
- Whether clicking Retry *before* physically power-cycling reproduces the
  expected immediate re-wedge (confirming the confirmation-gate is doing
  its job, not just that the UI shows a warning)
- Whether the durable `scanLog()` lines (auto-pause, retry attempt/outcome)
  are actually useful for post-incident diagnosis when read from a real
  `~/.bloom/logs/graviscan-*.log` file
- Real-world timing: does the accepted "in-flight row falls back to its
  90s timeout" tradeoff (design.md Decision 1) actually happen often
  enough on real hardware to matter, or is it as rare as the backtest
  suggests?

### Bench tests

- [ ] Deliberately induce a `sane_start_invalid` wedge (per the known
      root-cause conditions in issue #228) during an active interval
      session — confirm the scanner stops immediately and the banner
      appears within a few seconds
- [ ] Confirm the Slack alert and the in-app banner both fire from the
      same wedge (no divergence between the two notification paths)
- [ ] Confirm other scanners in the same session keep running unaffected
      while one is auto-paused
- [ ] Dismiss the banner, then physically power-cycle the wedged scanner,
[...]

Issue #243, opened 2026-05-26
upsertScannerRow created fresh UUIDs instead of UPDATING existing rows on Reset USB + Detect — DB/coordinator state divergence + duplicate-spawn race


eberrigan:
## Summary

During PR #237 rig validation on 2026-05-26, after a clean .deb deploy + Reset USB + Detect Scanners cycle, the GraviScanner DB table ended up with 5 fresh-UUID rows while the live \`scan_worker\` subprocesses were still tracked by the ScanCoordinator under a completely DIFFERENT set of UUIDs. The disable-not-delete policy from #230 was not honored — instead of UPDATING the existing rows via \`upsertScannerRow\`'s usb_port match, new rows appeared.

Symptoms observed:

- Duplicate-spawn race for one scanner (f6586a87) → one of two subprocesses kept losing the SANE \`open()\` race (\`Device busy\` × 3 attempts) and exited code 1, surfacing as a red status in the Configure Scanner UI
- DB and coordinator state diverged: live workers tracked by IDs not present in the GraviScanner table
- UI papers over divergence via \`usb_port\` join, so end-users see all-green
- Scans still complete successfully (proven by the 14-min × 2-cycle × 5-scanner × 4-plate run earlier in the same session, 40 plates uploaded)

## Timeline (2026-05-26 PDT)

1. ~13:00 — Old packaged app (May-14 install, PID 181158) gracefully stopped. DB at this point had scanner UUIDs \`f6586a87, 91832157, 9b46abac, c7536d45, ebb9a629\`.
2. ~14:13 — New .deb (built from \`bb97fc3\`, fix/v600-wedge-followups branch) installed via \`sudo dpkg -i\` (#226 workaround).
3. ~15:02 — New app launched via desktop shortcut. Auto-init spawned 5 \`scan_worker\` subprocesses with the SAME 5 OLD UUIDs (read from existing DB rows). One scanner (f6586a87) hit a duplicate-spawn race:
   \`\`\`
   15:02:15.252 › [ScanCoordinator] Scanner f6586a87 ready
   ... (4 seconds later) ...
   15:02:19.310 › [ScanCoordinator] Scanner f6586a87 init failed: Failed to open device after 3 attempts
   15:02:19.375 › [ScannerSubprocess:f6586a87] Exited: code=1
   \`\`\`
4. User clicked **Reset USB** button (Configure Scanner page) → didn't restore f6586a87.
[...]

PR #285, opened 2026-08-04
fix(graviscan): retryScanner() reports failure when respawned worker silently fails to come online

### LINKED ISSUE #283: GraviScan: retryScanner() reports success even when the respawned worker silently fails to come online
## Context

PR #277 (`add-graviscan-wedge-response-ui`, roadmap Tier 3, merged
2026-08-04) added `retryScanner()` (`src/main/graviscan/session-handlers.ts`),
which resolves `{success: true}` as soon as `coordinator.addScanner(...)`
resolves without throwing. This was an explicit, named "accepted
limitation" in the proposal's `design.md` (Decision 7's closing section),
signed off during the original 5-agent adversarial proposal review.

During PR #277's own post-implementation review round, two independent
reviewers (security, scientific rigor) re-flagged this as more severe
than "accepted": `ScanCoordinator.addScanner()`/`spawnSingleScanner()`
deliberately never throws on spawn failure — errors surface only via a
separate `scanner-init-status` event, recorded in `initErrors`. So a
respawned worker that silently fails to come online still produces
`retryScanner() -> {success: true}`, and `WedgeBanner.tsx`'s
`handleConfirmRetry` dismisses the banner entry on that success — telling
the operator the scanner recovered when it may still be dead. For a
continuous, multi-day, often-unattended time-lapse run (the primary use
case #244 was filed about), a scanner silently and permanently dropped
from rotation with the operator believing it's fixed is close to the
exact "permanent data loss, no recovery" failure mode this whole feature
exists to prevent — it just moves the failure from "the app doesn't
notice" to "the app tells the operator it worked when it didn't."

Decision made 2026-08-04: keep this deferred rather than reopening the
already-deliberated proposal scope with an unplanned fast-follow. Filing
this issue so the concern is tracked rather than silently dropped a
second time, per this repo's established pattern (see #279, #281).

## Suggested fix (not required to be built exactly this way)

`startScan()` already has the pattern needed for this, one function away:
its "final-review fix #3" (`session-handlers.ts`, `startScan()`) calls
`coordinator.getScannerStatuses()` — a synchronous poll of coordinator
state (`ready | starting | error | dead` per scannerId), not an
event-correlation-with-timeout — after `initialize()` resolves, and
returns failure if the scanner didn't actually come online. `retryScanner()`
never got the equivalent check. The fix is likely a small, well-precedented
addition:

```ts
await coordinator.addScanner({ scannerId, saneName, plates: [] });

const status = coordinator
  .getScannerStatuses()
  .find((s) => s.scannerId === scannerId);
if (!status || status.status !== 'ready') {
  return {
    success: false,
    error: status?.error ?? `Scanner ${scannerId} did not come online after retry`,
  };
}
return { success: true };
```

Should be verified (not assumed) that `addScanner()` fully settles a
scanner's status (ready or error) by the time it resolves, the same way
`initialize()` does for `startScan()` — this is the load-bearing
assumption the existing `startScan()` pattern already relies on, so it's
likely true, but worth confirming against `spawnSingleScanner()`'s actual
implementation before reusing the pattern here.

## Scope

- `src/main/graviscan/session-handlers.ts`'s `retryScanner()`
- `tests/unit/graviscan/session-handlers.test.ts`'s `retryScanner` suite
  (add a case: `addScanner()` resolves, but `getScannerStatuses()` still
  shows `error`/`dead` for that scannerId — assert `{success: false}`)
- No renderer changes expected — `WedgeBanner.tsx` already handles
  `{success: false, error}` correctly; it just never receives it for this
  failure mode today.

eberrigan:
## Summary

`retryScanner()` resolved `{ success: true }` as soon as `coordinator.addScanner(...)` resolved without throwing, but `addScanner()`/`spawnSingleScanner()` deliberately never throw on spawn failure — a respawned worker could silently stay dead while the operator was told it recovered. Fixes #283.

## Changes

- `retryScanner()` (`src/main/graviscan/session-handlers.ts`) now checks `coordinator.getScannerStatuses()` immediately after `addScanner()` resolves and returns `{ success: false, error }` if the scanner isn't reported `'ready'` — mirroring `startScan()`'s existing `hasWorker()` check after `initialize()`.
- All three `retryScanner()` `scanLog()` lines (success, caught-rejection failure, new status-check failure) now include `session_id`/`cycle_number`, matching the precedent set by `WedgeDetector`'s auto-pause log line.
- Test-only: the local mock `ScanCoordinatorLike` interface and `createMockCoordinator()` in `tests/unit/graviscan/session-handlers.test.ts` gain a `getScannerStatuses` member.
- No renderer changes — `WedgeBanner.tsx` already handles `{ success: false, error }` correctly.

## OpenSpec Change

- Change ID: `fix-graviscan-retry-scanner-status-check`
- Affected capability: `scanning`
- Delta type: MODIFIED (`GraviScan Retry-Scanner Action` requirement — full text carried forward, one scenario updated, one new scenario added)
- All `tasks.md` items complete: yes
- Went through 4 adversarial review rounds before implementation (5 parallel specialist reviewers in round 1, plus 3 focused follow-ups) — see `openspec/changes/fix-graviscan-retry-scanner-status-check/design.md` for the full settle-timing verification and review history.

### TypeScript Tests

- [x] Unit tests pass (`npx vitest run tests/unit/graviscan/session-handlers.test.ts` — 33/33, including 3 new tests for the silent-failure case)
- [x] Unit tests added for new functionality (error status w/ message, missing-from-list, dead-with-no-error-field; plus a happy-path assertion that `getScannerStatuses()` is actually called on success)
- [ ] Coverage threshold check — not run separately; existing file coverage pattern unchanged (mocked-coordinator unit tests, matching `startScan()`'s existing pyramid)

### Python Tests

- [ ] N/A — no Python changes

### Integration Tests

- [ ] N/A — no IPC/camera/DAQ/scanner/database changes; `graviscan:retry-scanner`'s IPC wiring itself is untouched

### E2E Tests

- [ ] N/A — no existing E2E test references `retryScanner`/`WedgeBanner` (verified via grep); this matches the existing test pyramid for this code (fully mocked-coordinator unit tests, same as `startScan()`'s analogous check)

## Build & Packaging

- [ ] Packaging — not run; this is a pure main-process logic change with no packaging-relevant surface

## Linting & Formatting

- [x] Prettier formatting applied (`npx prettier --check` on all changed files)

## Database Changes

- [x] No database changes

- [ ] CI Linux/macOS/Windows — will confirm once CI runs on this PR
- [x] No platform-specific assumptions — pure Node/TS logic, no path/OS branching

## Breaking Changes

- [x] No breaking changes — `retryScanner()`'s external `{ success, error }` IPC contract shape is unchanged; only when `success: false` is returned changes (now also on a previously-silent respawn failure)

## Related Issues

Closes #283
Related to #244, #240 (original wedge-detection issues), #277 (merged — introduced `retryScanner()`/`WedgeBanner`)
Out of scope, not addressed here: #279 (manual rig verification), #281 (three other pre-existing coordinator/subprocess risks), #243 (coordinator/DB scannerId join-key divergence — this fix relies on the same join key `retryScanner()` already used pre-fix)

## Reviewer Notes

The load-bearing assumption (that `addScanner()`'s promise never resolves before `spawnSingleScanner()` has fully settled internal state, on both the idle-path and mid-cycle-queued-path) is verified in detail in `design.md` Decision 1, independently re-derived from `scan-coordinator.ts`/`scanner-subprocess.ts` by multiple reviewers across the review rounds — worth a careful look if anything about `ScanCoordinator`'s internal timing changes in the future, since this check's correctness depends on it.

Also disclosed (not fixed here, pre-existing, out of scope): on the mid-cycle queued respawn path — the path an operator's retry almost always exercises in practice — `retryScanner()`'s promise doesn't resolve until the next `cycle-complete` event, which can be up to a full scan interval. This fix makes the eventual answer accurate; it doesn't make it faster. See `design.md`'s Risks section.


eberrigan (commented):

## Review Summary

Ran this through the `/review-pr` subagent team (5 lenses: code quality, testing/TDD, scientific rigor & UX, security & cross-platform, behavioural correctness) across 2 rounds. Round 1 found one real, well-corroborated defect and closed a test gap; round 2 verified both fixes and found nothing new. No blocking issues remain.

## Round 1 finding (fixed in e17d8fc)

**Dead `cycle_number` field in the new log lines.** Three independent reviewer lenses (code quality, testing, scientific rigor) independently caught that `session.currentCycle` — added to `retryScanner()`'s new `scanLog()` lines by analogy with `WedgeDetector`'s auto-pause log line — is set once to `0` in `startScan()` and never reassigned anywhere in `src/`. `ScanCoordinator` does track a real, live per-cycle counter internally, but nothing copies it back into session state, so every retry-outcome log line would have silently printed `cycle=0` for the life of every session — worse than omitting the field, since it's exactly the kind of confidently-wrong diagnostic data a researcher might later trust to reconstruct "scanner X was dead from cycle N onward."

Fix: dropped `cycle=` from all three log lines, kept `session=${session.sessionId}` (genuinely constant-correct for a session's lifetime). Wiring a real live cycle count through would require listening for coordinator cycle events in `session-handlers.ts` — a larger change, correctly out of scope for this narrow fix. Also closed a related test gap two reviewers flagged: the mock session now carries a `sessionId`, and three tests assert the log lines actually contain it (previously nothing tested the new log content at all).

## Round 2 verification

Independently re-ran `npx vitest run tests/unit/graviscan/session-handlers.test.ts` (33/33 pass), `npx tsc --noEmit`, `npm run lint`, and `npx openspec validate --strict` — all clean. Confirmed no stray leftover `currentCycle`/`cycle_number` references anywhere in the OpenSpec change docs or CHANGELOG. Confirmed via CI that `Test - Integration (windows-latest/macos-latest)`'s round-1 failures were genuinely pre-existing flakiness, not caused by this PR — they're green on the latest commit.

## Other findings from round 1 (verified pre-existing, not this PR's to fix)

- `Lint - Node.js` fails on `prettier --check` for `openspec/specs/scanning/spec.md` — confirmed via `git diff` that this PR never touches that file, and confirmed the same failure exists on `main`'s tip directly. Pre-existing drift, unrelated.
- The disclosed-but-unfixed mid-cycle-queued latency hazard (retry can hang up to a full scan interval before resolving) is a pre-existing behavior this PR's own analysis surfaced but correctly doesn't attempt to fix — same latency existed before, just with a wrong answer at the end; now it's the same latency with a correct one. `design.md`'s Risks section discloses it for a future follow-up.
- No security/cross-platform findings — diff is confined to main-process business logic with no new IPC handlers, file paths, or trust boundaries.

## Suggestions (optional, not blocking)

- Operator-facing error text (`status.error`, a raw SANE/subprocess message) can be cryptic for a non-programmer lab tech — pre-existing pattern across the whole function, not introduced or worsened here; worth a future UX pass.
- The `'starting'` status value is untested (structurally near-impossible to observe at the point `retryScanner()` reads it, per the settle-timing analysis in `design.md` — reasonable to leave as-is).

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), 2 rounds*

Issue #162, opened 2026-04-01
GraviScan: QR verification query not scoped to experiment and wave


blm3886:
## Bug

The `verify-plates` handler queries `GraviPlateSectionMapping` globally — it matches QR codes across all experiments and waves. This causes false `needs_review` flags when the same QR codes appear in multiple experiments or waves.

Additionally, plate_id comparison is case-sensitive — `plate_2` and `Plate_2` from the same metadata are treated as different plates.

### Example from production logs

Grid 00 detected 4 QR codes, but the DB lookup returned mappings from multiple plates across experiments:
```
{"plate_1":["D7Q3Z8R2M5LT",...], "plate_3":["N8P4Z2Y6T1KW",...]}
```

Grid 01 shows case mismatch: `"plate_2"` and `"Plate_2"` as separate entries.

## Expected

QR code lookup should be scoped to:
- The current experiment's accession (metadata file)
- The current wave (if applicable)

plate_id comparison should be case-insensitive.

## Where

- `src/main/graviscan-handlers.ts` — scope `findMany` query to experiment + wave
- `src/renderer/hooks/useScanSession.ts` — pass `experimentId` to `verifyPlates()`
- `src/main/preload.ts` + `src/types/electron.d.ts` — update type signatures

eberrigan:
Confirmed merged: `src/main/graviscan/verify-plates.ts` requires a non-empty `experimentId` and scopes queries to `(experiment_id, wave_number)`. Shipped across PRs #270/#278/#289 (2026-07-30 to 2026-08-25). Verified as part of the GraviScan production-cutover roadmap (docs/superpowers/plans/2026-09-02-graviscan-production-cutover-roadmap.md, Tier 0).

PR #289, opened 2026-08-05
GraviScan: add Core Scan-Operation screen (roadmap Tier 4)

eberrigan:
## Summary

Implements Tier 4 of the GraviScan renderer roadmap (`docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md`): the Capture Scan screen, the operator-facing UI for actually running a GraviScan scan session. Full OpenSpec proposal at `openspec/changes/add-graviscan-capture-scan-screen/` (design.md, proposal.md, tasks.md, spec deltas), converged after 4 rounds of `openspec-review`.

- **Schema**: `GraviScanPlateAssignment.wave_number` (fixes a real data-loss bug in plate-assignment overrides across waves), `GraviScan`'s new unique constraint (idempotent per-job persistence).
- **Backend**: wave-scoped `graviscanPlateAssignments` handlers, idempotent `graviscansCreate`, wave-scoped `verify-plates.ts` (read + write, issue #162), new preload wiring for `verifyPlates` + verify-* events.
- **Renderer hooks**: `useScannerStatus`, `useWaveNumber`, `usePlateAssignments` (schema-backed wave scoping, auto-fill with override), `useContinuousMode`, `useScanSession` (reducer-based session state, backend persistence wiring, abnormal-termination marker, wedge-blocked start), `useTestScan`.
- **Components**: `CadenceWarningBanner`, `QRVerificationBanner` (graded severity), `ScanFormSection`, `ScanControlSection`, `ScannerStatusPanel`.
- **Screen + routing**: `GraviScan.tsx` composes all of the above; wired into `App.tsx`'s `capture-scan` route and `Layout.tsx`'s nav for graviscan mode (previously an unreachable route — `WorkflowSteps.tsx`'s existing step 5 now actually navigates there instead of falling through to Home).
- **Shared `WedgeContext`**: lifts `useWedgeEvents()` out of `WedgeBanner.tsx` into `Layout.tsx` so wedge state is consistent across navigation, fixing a gap where a screen mounting its own independent subscription would miss a wedge that occurred while the operator was elsewhere.

Two bugs found and fixed during implementation (not present in the accepted design, found via testing against the real preload contract):
1. `register-handlers.ts`'s `wrapHandler()` wraps `startScan`/`cancelScan`/`getScanStatus`/`getOutputDir`/`getConfig` in a `{success, data}` envelope on their normal path; the new hooks initially read these as flat objects (would have silently misbehaved against the real app despite passing unit tests with incorrectly-shaped mocks). Added a shared `unwrapGraviResult()` helper and fixed all call sites — confirmed against `ConfigureScanner.tsx`'s existing `result.data?.isActive` usage and the real E2E suite.
2. A pre-existing ESLint `no-restricted-imports` glob (`**/graviscan/**`) intended to block shared code from importing `src/main/graviscan/*` also matched the unrelated, legitimately-named `src/renderer/components/graviscan/` directory this tier's own design specifies. Scoped the pattern to `**/main/graviscan/**` to match the override list's own existing scoping — verified the real architecture boundary is still enforced.

## Test plan

- [x] `npm run lint && npm run format:check`
- [x] `npx tsc --noEmit`
- [x] `npm run test:unit` (full suite) — 1408 passed; 5 pre-existing failures unrelated to this change (Windows path-separator assertion mismatches in `config-store.test.ts`, `image-uploader.test.ts` ×3, `scan-coordinator.test.ts`)
- [x] `npm run test:e2e -- tests/e2e/graviscan-ipc.e2e.ts` (full file) — 15/16 passed; the one failure (USB-reset mock timing) is a pre-existing, self-documented timing-sensitive test unrelated to this change
- [x] `./scripts/verify-migrations.sh`-equivalent parity check (no `sqlite3` CLI available locally; verified via a `better-sqlite3` schema diff instead) + `npm run test:db-upgrade` (34 passed)
- [x] `openspec validate add-graviscan-capture-scan-screen --strict`
- [ ] Manual live-Electron smoke test of the full happy path (select experiment/wave → assign plates → start continuous scan → observe progress → QR verification banner → restore-on-navigation → wave round-trip → force-quit/relaunch abnormal-termination banner) — not run as part of this PR; needs a human at a real machine with real scanners/mock hardware.


eberrigan (commented):
> **Round 1 review: findings confirmed and fixed** (posted as comment — cannot self-approve/request-changes on own PR)

## Review Summary

Ran a 5-lens adversarial review (code quality/architecture, testing/TDD, scientific rigor/UX, security/cross-platform, behavioural correctness) against this PR. All 5 subagents found real, confirmed issues — most severe: a field-name mismatch that silently broke QR verification and DB write integrity for every real scan. All findings below have been fixed and pushed (commit a2c4a49), each with a regression test.

- **Field-name mismatch breaks QR verification + DB integrity for every real scan.** `scan-coordinator.ts` only ever forwards the image location as `path` and the cycle as `cycle_number` (never camelCase `imagePath`/`cycleNumber` duplicates, unlike `scanner_id`/`plate_index`). `useScanSession`/`useTestScan` read the nonexistent camelCase fields, so every captured scan's DB row got `path: ''` (breaking image lookup) and `cycle_number: null` (silently disabling `graviscansCreate`'s idempotency upsert), and QR verification never ran at all (every plate filtered out by the always-empty `imagePath` check). Fixed with dual-casing resolvers matching the existing `resolveScannerId`/`resolvePlateIndex` pattern.
- **Stale-ref race could permanently stick a session in "scanning".** The "are all jobs done" check read a `stateRef` that only refreshes after a render commits; two `scan-complete` events for the last two remaining jobs delivered before that commit would both see the same stale snapshot and neither would detect completion. Replaced with a synchronously-updated ref-based completion tracker (immune to render timing).
- **No re-entrancy guard on `startScan()`/`testAllScanners()`.** A fast double-click ran two overlapping async setup chains, corrupting shared refs and leaking the first call's event listeners permanently. Both now guard against concurrent invocation.
- **Both functions could permanently lock up if `getOutputDir()` rejected** (vs. resolving with a failure) instead of just returning an error — now wrapped in try/catch/finally.

## Important Issues (all fixed)

- `graviscanSessions.create()` and `database.graviscans.create()` failures were silently swallowed — both now surface an error rather than degrading idempotency/traceability invisibly.
- `usePlateAssignments`'s persisted writes (auto-fill write-through and operator edits) were fire-and-forget — a failed save was indistinguishable from a successful one until the next remount. Added a `saveError` field, surfaced in `ScanFormSection`.
- `graviscanPlateAssignmentsList`/`UpsertMany` and `graviscansCreate`'s `wave_number` handling accepted any `typeof === 'number'` value (including `NaN`/negative/non-integer) instead of validating via the already-exported `isValidWaveNumber()`, inconsistent with `verify-plates.ts`'s own boundary. Fixed.
- No unmount/cleanup tests existed for `useScanSession`/`useScannerStatus`'s subscriptions — noted, not yet added (lower severity; the cleanup code itself was read and looks correct).

## Suggestions (not applied, low priority / accepted as-is)

- Renderer-side `output_path` construction has no containment validation before reaching the main process (pre-existing pattern this PR extends with DB-influenced values for the first time; practical exploitability assessed as low today given current UI input constraints — recommend as a follow-up, not blocking).
- A few `any`-casts in `useScanSession.ts`/`useTestScan.ts` are broader than strictly necessary.
- Minor: a raw `setTimeout` instead of `waitFor` in one `usePlateAssignments.test.ts` assertion.

All fixes verified: `npm run lint`, `npx tsc --noEmit`, and the full unit suite (1425 passed, 5 pre-existing unrelated failures) all green after the fixes.

---

eberrigan (commented):
> **Round 2 review: verified round-1 fixes, found and fixed one new regression** (posted as comment — own PR)

Ran a targeted adversarial re-check of the round-1 fix commit (a2c4a49) specifically — verifying each claimed fix and looking for anything the fix itself broke.

## Verified fixed (all 7 round-1 findings)

Confirmed correct with direct code inspection: the imagePath/cycle_number field resolution, the completedKeysRef/allDoneFiredRef reset-on-START and seed-on-RESTORE logic, the re-entrancy guards' `finally` coverage across every early-return path, the try/catch/finally wrapping the whole async chain (not just the outputDir call), the START-then-ERROR dispatch ordering for session-create failures, `usePlateAssignments`'s `saveError` wiring to both write sites (with correct clear-on-success), and the `isValidWaveNumber` validation in all three `database-handlers.ts` functions (correctly still accepting an omitted waveNumber while rejecting a provided invalid one).

## New regression found and fixed

The completedKeysRef fix for the stale-ref race (round 1) only added a job's key to the tracker in `onScanComplete` — `onScanError` dispatches `JOB_ERROR` (which does remove the job from `pendingJobs`) but never touched `completedKeysRef`. Consequence: a single-mode (non-continuous) session where any one job errors and the rest complete normally could never reach `completedKeysRef.size >= totalJobs` — `SCAN_ENDED`/`finishSession` would never fire, leaving the session permanently stuck showing `isScanning: true`, with no session-complete DB write and no QR verification ever run. Fixed (commit 557e0dc) by extracting the "mark job accounted for, check all-done" logic into a shared `markJobAccountedFor()` called from both `onScanComplete` and `onScanError`, with a new regression test covering the mixed complete+error scenario.

Verified: `npm run lint`, `npx tsc --noEmit`, full unit suite (1426 passed, same 5 pre-existing unrelated failures) all green.

---
*Round 2 — verification pass on the round-1 fix commit*

eberrigan (commented):
> **Round 3 review: verified round-2's fix, found and fixed one more guard gap** (posted as comment — own PR)

Targeted re-check of commit 557e0dc (the round-2 fix). All 7 verification points confirmed correct: closure freshness across the effect's dependency chain, the 2-job and 3-job mixed complete/error scenarios, no conflict between the reducer's `pendingJobs` pruning and the ref-based `completedKeysRef` bookkeeping, no leftover old-style "all done" check, Set-based idempotency for duplicate error events, and confirmed the new regression test would genuinely have failed on the pre-fix commit.

## New gap found and fixed

`onScanComplete` already guards against an unknown job key (`if (!job) return;`) before touching the shared completion bookkeeping, but `onScanError` had no equivalent guard — before round 2's fix this asymmetry was inert (only the guarded path fed the bookkeeping), but once both handlers started feeding the same `completedKeysRef`, a stray or mismatched `scan-error` event (a leftover from a prior session, a hardware echo with a wrong plate index) could inflate the completion count past the real job total, risking a **premature** `finishSession(false)` while a real job was still in flight. Fixed (commit 332850b) by adding the same `if (!(key in jobTemplateRef.current)) return;` guard to `onScanError`, mirroring `onScanComplete`, with a regression test proving a stray event no longer ends the session early.

Verified: `npm run lint`, `npx tsc --noEmit`, full unit suite (1427 passed, same 5 pre-existing unrelated failures) all green.

## Convergence assessment

Three review cycles: round 1 found multiple severe, independent bugs; round 2 found one regression introduced by round 1's own fix; round 3 found one narrower gap in round 2's own fix (a missing guard mirroring an already-established pattern in the same file). The finding severity and scope has been shrinking each round. Treating this as converged — no further review cycles planned unless something else surfaces.

---
*Round 3 — verification pass on the round-2 fix commit*

eberrigan:
## Live smoke-test round: 4 more fixes

Manual smoke testing (task 16.6) surfaced several real bugs the automated `/review-pr` rounds missed. Fixed and pushed (commits `012e9d2`, `6ac0d31`, `14c1dd3`):

1. **Continuous-mode data loss (critical).** `useScanSession.ts` built `output_path` without a `_cy` placeholder, so `scan-coordinator.ts`'s per-cycle filename rewrite silently no-op'd — every cycle overwrote the previous one's image at the identical path. Confirmed via the rig log: the same filename repeating across cycles 1–4. Fixed by including a `_cy1_` token in the built path, matching the existing `_cy\d+_` rewrite regex.
2. **Silent re-start.** Clicking Start while a scan was already running/starting did nothing visible. Now dispatches "A scan is already in progress."
3. **Stale abnormal-termination banner.** The banner was only ever set, never cleared, so it persisted across a later clean scan. Now cleared on a fresh successful `startScan()` and wherever the underlying localStorage marker is cleared.
4. **Unstyled Start/Cancel/Test Scan buttons** — had no styling classes at all. Styled to match existing app button conventions (`CaptureScan.tsx`, `WedgeBanner.tsx`).
5. **Invisible plate-grid column headers** — barcode/transplant-date/custom-note inputs relied on `sr-only`/`aria-label` only, with no visible indication to sighted users of what to type. Added a visible header row.
6. **No cross-navigation session persistence** — experiment/phenotyper/wave were local state, lost on navigating away and back. Wired into the existing `window.electron.session` mechanism already used by `CaptureScan.tsx`.
7. **Continuous-mode sessions ending after cycle 1 (found during investigation, confirmed real).** The "all done" job-completion counter didn't distinguish single-mode from continuous-mode: as soon as one cycle's jobs were all accounted for, it fired full session completion (DB row marked complete, QR verification run, UI showed "not scanning") — while the backend coordinator kept scanning cycles 2..N independently. This also meant a user could click Start again mid-session, since the re-entrancy guard reads `isScanning`, and the real `interval-complete` event later re-ran the same completion sequence a second time. Continuous mode now relies solely on `interval-complete` to end the session, with a new `CYCLE_ADVANCE` reducer action resetting per-cycle job state on each `cycle-complete`.

All fixes have regression tests (`tests/unit/hooks/useScanSession.test.ts`, `tests/unit/pages/GraviScan.test.tsx`). Full `tsc --noEmit` + lint + unit suite clean (pre-existing, unrelated failures only: Windows path-separator assertions and flaky `AccessionForm` timing tests, same as documented in tasks.md 16.2).

**Follow-up filed, not in this PR:** duplicate plant-barcode entries within the same wave aren't currently blocked. There's an unused backend handler (`graviscansCheckBarcodeUniqueInWave`) that could be wired up for this later — out of scope per the original design.md/tasks.md.

Not merging — resuming the manual smoke-test walkthrough (task 16.6) next.

eberrigan (commented):

## Review Summary

Reviewed via 5 parallel subagents (Code Quality/Architecture, Testing/TDD, Scientific Rigor/Metadata/UX, Security/Cross-Platform, Behavioural Correctness) against the full diff plus focused reading of the ~16 commits since round-1 review. Overall: strong TDD discipline and documentation (each recent fix names its root cause, cites a confirmed-red test, and is recorded as a numbered Decision in `design.md`). However, this pass surfaced several genuine data-integrity bugs in the core session hook that I independently verified against the code (not just trusting subagent claims) — these should block merge.

## Blocking Issues

1. **Session context (experiment/phenotyper/wave) isn't frozen once a scan starts.** Verified: `GraviScan.tsx` has no `disabled` wired to `scanSession.isScanning` on `ExperimentChooser`/`PhenotyperChooser`/the Wave input. `useScanSession.ts`'s `recordCompletedJob`/`runVerification` both read `contextRef.current` — a ref mirrored live on every render — not a value frozen at session start. `clearAbnormalMarker` similarly closes over the *current* `experimentId`/`waveNumber` (via `useCallback` deps), while `startScan()`'s marker write used its *own* closure at session-start time. Net effect: if an operator switches experiment/wave while a scan is running, (a) completed-job DB writes and QR verification for still-in-flight jobs get attributed to the *new* experiment/wave, not the one the session actually started under, and (b) `finishSession`'s cleanup can clear the wrong wave's abnormal-termination marker — leaving the real session's marker stuck (a false "didn't finish cleanly" banner later) or clearing an unrelated wave's real marker. Fix: snapshot the session's experiment/phenotyper/wave once at `startScan()` and use that snapshot (not live refs) for the lifetime of the session; disable the selectors while `isScanning`.

2. **`completedJobsRef` is never restored on remount.** Verified: the mount-time `RESTORE` path rebuilds `jobTemplateRef`/`completedKeysRef` from the backend's `status.jobs`, but never touches `completedJobsRef` — the *only* thing `runVerification()` reads at session end. Navigate away and back mid-scan, and every plate that completed *before* the remount is silently excluded from QR verification, with no error or "partial verification" indicator.

3. **Wrong resolution persisted to the `GraviScan` DB record.** Verified: `recordCompletedJob` writes `resolution: ctx.resolution` — the *requested* DPI from the form — never `achieved_resolution` from the scan-complete event. `database-handlers.ts:128` has an explicit, pre-existing comment warning that this exact field must come from `achieved_resolution` or "the #232 fix never reaches the queryable database record." This silently misrepresents actual scan resolution in the DB whenever hardware doesn't hit the requested DPI.

4. **A known write-race was found but shipped untracked.** `persistPosition()`'s `upsertMany()` call is fire-and-forget with no navigation guard. Commit `921aa34` explicitly identifies this as the likely explanation for a live bug report (a plate-assignment edit not persisting) but ships no fix, no regression test, and no filed issue — only a commit-message note. Please fix it, or at minimum file and link a tracked GitHub issue before merge.

5. **CI wasn't fully green at review time** — several jobs (`Test - Make (Windows)`, `Test - E2E Dev Build` on ubuntu/macOS/windows) were still pending. These real-Electron E2E jobs are exactly where the mock-shape/timing bugs already found elsewhere in this PR would surface — please confirm full green before merge.

## Important Issues

- Three new files (`useScanSession.ts`, `useTestScan.ts`, `ScanControlSection.tsx`) each open with a blanket `/* eslint-disable @typescript-eslint/no-explicit-any */` to license `(window as any).electron.gravi.*`, even though `window.electron` is already fully typed via `declare global` in `electron.d.ts`. This throws away real typing and forces manual re-casts elsewhere (e.g. `useScanSession.ts:391`) that the typed API would give for free.
- `INTERVAL_WAITING`/`INTERVAL_RESUMED` (the new coordinatorState actions) have no guard against a session that has already ended — a stray/late event after `SCAN_ENDED` could flip `coordinatorState` back to `'waiting'`/`'scanning'`, and `ScanControlSection.tsx`'s "Waiting for next cycle..." banner has no `isScanning` gate to catch it.
- `SCAN_ENDED` doesn't clear `pendingJobs`/`progressByScanner` the way `CANCELLED` does — an asymmetry that risks `isScanning: false` alongside stale non-empty `pendingJobs`.
- `QRVerificationBanner.tsx` has no case for `status: 'swapped'` — a real, audit-trailed plate-swap auto-correction falls through to the plain green "every plate verified correctly" banner, indistinguishable from a genuinely clean run.
- Interval validation is displayed but not enforced: `ScanControlSection.tsx` shows `continuousMode.validate()`'s error banner, but "Start Scan" isn't gated on it — an operator can start anyway with an invalid interval forwarded to the backend unchanged.
- The `interval-waiting`/`coordinatorState` fix has zero integration/E2E coverage of the *real* event path through `scan-coordinator.ts` — only unit-tested via hand-fired mock events, which is precisely the "wired on paper, never driven by a live event" bug class this PR repeatedly found and fixed elsewhere.
- Write-side `output_path` construction (`useScanSession.ts`, `useTestScan.ts`) has no path-containment check, unlike every read-side path handler this PR touches (low risk today since inputs are UUID/enum-constrained, but no defense-in-depth if that ever changes); also uses string concatenation with a hardcoded `/` instead of `path.join()`.
- A new `.catch(() => {})` (the abnormal-marker check effect) swallows silently with no logging, inconsistent with sibling catches in the same file that dispatch a visible error.
- `transplantDate`'s Date-object-vs-string mismatch (already tracked as #315) has a wider blast radius than "displays blank": since `fieldsEqual()` compares via `===`, any plate position with a transplant date is *permanently* misclassified as operator-overridden on every load, silently defeating the metadata auto-fill re-sync mechanism (Decision 3) entirely. Worth updating #315 to capture this, not filing a duplicate.

## Suggestions

- UI-only fixes (button/banner styling, grid alignment) shipped with no test — a lightweight assertion on expected Tailwind classes would catch the original regression class cheaply.
- Name the 300ms session-save debounce (`GraviScan.tsx`) as a constant.
- Add a timeout guard around `getScanStatus()` in the restore/marker-check effects — a hang currently leaves state silently unresolved.
- `graviscansCreate` doesn't validate the `path` field itself before persisting.
- The mount-time restore effect and the Decision-10 abnormal-marker effect both independently call `getScanStatus()` — a documented trade-off, but worth consolidating.

---

eberrigan (commented):
> **Verdict: REQUEST_CHANGES resolved — all blocking issues from this round fixed before posting** (posted as comment — cannot request changes/approve own PR)

## Review Summary (round 2 — reviewing the round-5 fixes themselves)

Ran the same 5-subagent team (`/review-pr`) against the fixes for the previous round's 4 major bugs + 7 important issues, rather than against the original implementation. All of round 5's original findings were independently re-confirmed as genuinely closed. This round's own fixes introduced their own issues, though — including one real security bug and a regression test that didn't prove what it claimed to. All findings below were fixed in this same pass, verified (`tsc --noEmit`, lint, `format:check`, full `test:unit` — 1668 passed, same 4-5 pre-existing/flaky failures, zero new), and pushed.

## Blocking Issues (fixed)

- **Prototype pollution** — `markScanJobRecorded` (`wiring.ts`) used a truthy bracket-access check (`scanSession?.jobs[key]`) on a renderer-controlled `key` against a plain object literal. A call with `key: '__proto__'` resolved through the prototype chain to `Object.prototype` itself and mutated it, polluting every plain object in the main process for the app's lifetime. Fixed with `Object.prototype.hasOwnProperty.call(...)`. Added regression tests proving the pollution occurred pre-fix (confirmed red, then green).
- **The write-race regression test (Section 25) didn't prove the fix** — it resolved the "slow" write synchronously immediately after the fresh mount rendered, before that mount's own load effect ever reached its `.list()` call, so it passed identically whether or not the guard existed. Rewritten to assert `listAssignments` hasn't been called yet (call-count comparison across the fresh mount) before resolving the write. Confirmed genuinely red against a temporarily-disabled guard, then green.
- **Write-vs-write race** — two rapid edits to the *same* plate position fired concurrent, unchained `upsertMany()` calls, and the write-tracking map only ever held the last-registered promise per key, so an older write settling after a newer one had no protection and no future visibility. Replaced `registerPendingWrite()` with `enqueueWrite()`, which now serializes writes to the same key instead of firing them concurrently. New test proves this with a controllable-resolver mock; confirmed red against a temporarily-reverted chain, then green. Also added a bounded 10s timeout to `awaitPendingWritesFor()` so a genuinely hung IPC call can't block a future mount forever.

## Accepted, narrower limitations — filed as follow-up issues rather than fixed here (per your decision)

- **#330** — if the renderer crashes/reloads in the gap between `recordCompletedJob()` resolving and `markJobRecorded()`'s IPC round-trip completing, that job stays `'pending'` on the backend forever; a restored single-shot session can never reach "all done." The DB write itself isn't lost — only session/verification bookkeeping.
- **#331** — the same write-loss risk the plate-assignment fix targets for navigate-away-and-back can also occur via a full app quit; `main.ts`'s `before-quit` handler doesn't await in-flight renderer IPC writes.
- **#332** — `path-containment.ts`'s missing-tail reappension was written for read-only callers; `start-scan`'s new use is its first writer, and it has a TOCTOU window that requires an attacker to already have an independent local filesystem-write primitive to exploit.

## Important Issues (fixed)

- `session-handlers.ts`'s inline job-status type was missing the `'recorded'` literal that the `markJobRecorded` wiring fix makes live for the first time — now imports the canonical `ScanSessionJob` type instead of a locally-duplicated, drifted one.
- `electron.d.ts`'s `startScan`/`markJobRecorded`/`cancelScan`/`getOutputDir` — the exact four calls the any-removal fix touched — were still typed `Promise<any>`, so that fix's stated goal didn't reach its own highest-risk call site. Fixed with hand-duplicated local interfaces (shared code can't import from `main/graviscan/*` per an existing ESLint rule).
- The new Windows containment test compared an un-resolved forward-slash path string against the mock's resolved (backslash-normalized) path, so it always reported "exists" on the first check and never actually exercised the missing-tail branch it claimed to. Fixed by resolving the comparison path up front, matching `ensure-dir`'s own correct pattern.
- `markJobRecorded`'s own swallowed `.catch(() => {})` now logs via `console.error`, matching the sibling fix from two commits earlier in the same round. Added a rejection test.
- `SCAN_ENDED`/`CANCELLED`'s now-identical reducer cases were left as two separate copies — exactly the "two copies that can silently drift apart" shape that caused the original asymmetry. Consolidated into one shared case.
- No operator-visible explanation for why the experiment/phenotyper/wave selectors lock during a scan — added a `title` tooltip.
- The swap-verification banner named the count but not that no action is needed — added "— no action needed" so it doesn't rely on the operator inferring that from the color alone.
- "Test Scan" wasn't gated the same way "Start Scan" now is — added `scanSession.isScanning` to its disabled condition.
- The start-scan containment loop had no guard against a malformed `params.scanners` shape (would throw an unhandled, internal-detail-leaking error instead of the uniform `{success:false, error}` shape). Added a defensive guard plus tests for malformed input, a multi-scanner/multi-plate escape case, and the `getOutputDir()`-failure branch.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), round 2 — reviewing the fixes for round 1's own findings.*

eberrigan (commented):

## Round 3 Review Summary (openspec-review + /review-pr, 10 subagents)

Third full review cycle, covering everything since round 2's synthesis: the `origin/main` merge (a large concurrent CylinderScan style/UX PR landed there), Decision 23/Section 33 (a real bug found during live manual smoke testing — the scanner-status panel stuck on "disconnected" throughout an active scan), and this round's own findings on that fix. Converged: all substantive findings fixed, verified, and pushed.

- **Race condition in `useScannerStatus.refresh()`** (Behavioural Correctness lens): rapid Start/Cancel/Start cycles, plus `GraviScan.tsx`'s own isScanning-triggered call, could have two `getScannerStatus()` IPC round-trips in flight with no guarantee of resolving in request order — an older response resolving after a newer one would silently revert the panel to stale status, reintroducing Decision 23's bug via concurrency. Fixed with a monotonic `latestRequestIdRef`; a response is applied only if no newer request has been issued since. Also exposed the mount effect's `cancelled` flag as dead code (never actually gated the state setters) — removed as redundant.
- **CI's `Lint - Node.js` was genuinely failing** on `format:check` for an archived doc pulled in via the merge — confirmed via the actual job log (not a local-checkout artifact). Same Prettier non-idempotency class as `dc7cff9`'s `tasks.md` fix. Fixed.
- **`openspec-review`'s own coverage gap**: task 33.1's test only proved `useScannerStatus.refresh()` works in isolation, never that `GraviScan.tsx`'s wiring actually calls it at the right time — and `GraviScan.test.tsx`'s mock had no `refresh` field, so a future test triggering the transition would've crashed with a TypeError instead of failing meaningfully. Fixed with a real `rerender()`-based regression test.

## Important Issues (fixed)

- **Accessibility regression pulled in by the merge**: `ExperimentChooser.tsx`/`PhenotyperChooser.tsx` (shared with GraviScan's Capture Scan screen) had their focus ring changed from `blue-500` (~3.68:1 contrast) to `lime-500` (~1.98:1) by PR #329's CylinderScan style pass. That PR's own `design.md` flagged the shade as "novel, not attested in the pilot" with its manual contrast check left unchecked — and lime-500 fails WCAG 2.1 SC 1.4.11's 3:1 minimum for focus indicators. Fixed by changing just the ring shade to `lime-700` (~5:1), keeping the intended lime accent without touching anything else PR #329 colored.

## Suggestions (not applied — optional)

- A regression test asserting `useScanSession`'s initial render always starts `isScanning: false` (the invariant Decision 23's ref-based mount-skip logic implicitly depends on) would make that cross-file dependency explicit rather than only reasoned about.
- The green/swap QR-verification banner path remains unverified live (needs wave-metadata-linking UI, unstarted Tier 5 work) — unit-tested, honestly disclosed in task 16.6, judged acceptable to merge without live verification, but worth requiring as an acceptance criterion when that Tier 5 UI ships.

## Verification

Full unit suite: 1767 passed, 7 failed across 4 files, all matching the already-documented pre-existing/flaky set (config-store, image-uploader x3, scan-coordinator's regex test, electron-cleanup.test.ts) — zero new failures. `tsc --noEmit`, lint, and `format:check` all clean. `openspec validate --strict` passing. CI is re-running now on the latest push.

---
*Review by Claude Code subagent teams — openspec-review (5 lenses) + /review-pr (5 lenses), round 3*

eberrigan (commented):

## Round 4 Review Summary (/review-pr, 5 subagents)

Run once CI was fully green (all 20 checks passing on commit e65f2a4). A genuinely fresh adversarial pass, not just re-confirming prior rounds. Found one real bug and several smaller hardening items; all fixed and pushed.

## Blocking Issues

None found this round.

## Important Issues (fixed)

- **Unhandled promise rejection in `useScannerStatus.refresh()`** (Behavioural Correctness lens): a genuine `getScannerStatus()` rejection (distinct from the already-handled `{success: false}` path) left `loading` stuck `true` forever and `scanners` stuck `[]` — which keeps the poll-while-`starting` effect from ever starting. No retry path short of an app restart — strictly worse than the PR #213 bug this hook exists to fix. Fixed with try/catch/finally, confirmed via a test that reproduced the actual unhandled rejection before the fix. Same shape fixed in `useWaveNumber.ts` for consistency.
- **`preload.ts`'s `graviAPI` had no `: GraviAPI` type annotation** (Code Quality lens), unlike every sibling API object in that file — the compiler never checked it against the interface renderer code trusts. Added; confirmed no existing drift.
- **`verify-plates.ts`'s swap-correction write is wave-unscoped if a future caller omits `waveNumber`** (Scientific Rigor lens) — dormant today (the one live caller always passes it), and `waveNumber` stays optional by design for backward compatibility with ~50 pre-wave-scoping callers, so this wasn't changed. Added an explicit warning comment at the write site instead.

## Suggestions (not applied — left for explicit decision)

- **`QRVerificationBanner.tsx` shows only aggregate pass/fail text, never which specific plate had an issue** — a real actionability gap for a lab technician who needs to know which physical plate to inspect. `QRVerifyPlateResult` already carries per-plate detail; this is a UI-scope question (how much to redesign the banner), not a one-line fix, so it's left for an explicit decision rather than expanded into unilaterally.
- `useScannerStatus.ts`'s `toScannerPanelState()` hardcodes several `ScannerPanelState` fields it has no data for (reused type from `ConfigureScanner.tsx`) — documented with a comment rather than forking the type.
- `ExperimentChooser.tsx`/`PhenotyperChooser.tsx` are ~95% duplicated — pre-existing, not this PR's debt, worth a follow-up extraction.

## Verification

Full unit suite: 1770 passed, 6 failed across 4 files, all matching the already-documented pre-existing/flaky set — zero new failures. `tsc --noEmit`, lint, and `format:check` all clean. `openspec validate --strict` passing. CI re-running now on the latest push (commit ecc5bfa).

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), round 4*

Issue #288, opened 2026-08-05
downloadImages CSV export can silently mismatch for wave-scoped GraviScan accessions


eberrigan:
## Background

Tier 5 (GraviScan Browse / Experiment Detail / Metadata UI,
`openspec/changes/add-graviscan-tier5-browse-metadata`) added wave-scoped
metadata linking (`GraviExperimentWaveMetadata`), letting different waves of
the same experiment link to different accession/metadata files via
`experiments.linkGraviMetadata`/`unlinkGraviMetadata`.

`downloadImages` (see `src/main/image-handlers.ts` around lines 501-505)
still exports a single CSV derived from the experiment's own
`accession_id` field, not from the per-wave link table. For an experiment
whose waves have been linked to metadata files that diverge from the
experiment-level default (or from each other), the exported CSV's
accession/plate mapping can silently mismatch what a given wave was
actually scanned against.

Tier 5 mitigated this with a UI-visible warning banner before download
(see `design.md` Decision 10 in the Tier 5 OpenSpec change, and the
"Mismatch warning before downloading a single diverged wave" / "...an
'All Waves' download..." scenarios in
`openspec/specs/ui-management-pages/spec.md`), but the underlying
`downloadImages` CSV-generation logic itself was explicitly left
unchanged — the fix was scoped to warning the user, not correcting the
export.

## Ask

Make `downloadImages`'s CSV export wave-aware: for each wave present in
the requested scan set, resolve the accession/plate mapping from that
wave's `GraviExperimentWaveMetadata` link (falling back to the
experiment-level default only for waves with no link), instead of always
using the experiment's single `accession_id`.

## References

- `openspec/changes/add-graviscan-tier5-browse-metadata/design.md` — Decision 10
- `src/main/image-handlers.ts:501-505`

PR #290, opened 2026-08-05
GraviScan: add Browse / Experiment Detail / Metadata UI (Tier 5)

### LINKED ISSUE #286: GraviScan: no UI path exists to create a GraviScan-typed Experiment
## Problem

`Experiment.experiment_type` defaults to `"cylinderscan"` (`prisma/schema.prisma:36`) and nothing in the renderer ever overrides it. `src/renderer/components/ExperimentForm.tsx`'s zod schema (`name`, `species`, `scientist_id`, `accession_id` only) never includes `experiment_type`, and its `onSubmit` calls `database.experiments.create` with a payload that omits the field entirely.

Confirmed via search: the only places that ever set `experiment_type: 'graviscan'` are test fixtures (`tests/unit/graviscan/database-handlers.test.ts`, `tests/e2e/renderer-database-ipc.e2e.ts`, `scripts/generate-db-fixtures.ts`). No open or closed issue tracks this (checked #133, #207, #164, and searched for "experiment_type" / "create experiment" / "graviscan experiment").

## Impact

Every experiment created through the app today is CylinderScan-typed, regardless of the active `scanner_mode`. This blocks the wave-scoped metadata-link feature (`experiments.linkGraviMetadata`, added in PR #278) end-to-end in normal use, since `linkGraviMetadata` requires `experiment_type === 'graviscan'` (`database-handlers.ts:953`) and no real (non-fixture) experiment satisfies that today.

Note: this is distinct from #275, which is about `Accessions` lacking a file-type discriminator — that's a different table and a different gap.

## Proposed fix

Make `ExperimentForm.tsx` mode-aware: when creating an experiment while `scanner_mode === 'graviscan'`, set `experiment_type: 'graviscan'` on the create payload (backend handler already accepts the field with no restriction — `database-handlers.ts` `experiments.create`, lines ~1137-1159 — so no backend change needed).

Scoped into **Tier 5** of the GraviScan renderer roadmap (`docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md`) as a blocking prerequisite for that tier's wave-scoped metadata-link UI.

eberrigan:
## Summary

Implements the GraviScan roadmap's Tier 5 (Browse / Experiment Detail / Metadata UI) — see `docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md` and `openspec/changes/add-graviscan-tier5-browse-metadata/`.

- `BrowseGraviScans`: filterable, paginated list of graviscan experiments with a wave/accession mismatch warning, "Backup to Box" (rclone-unavailable / partial-failure messaging), links through to Experiment Detail
- `ExperimentDetail`: scan/file listing with scanner/wave filters, resizable columns, file preview, inline wave-scoped metadata link/unlink (durable audit log line on both actions)
- `Metadata`: spreadsheet upload (multi-sheet, column auto-mapping, per-row validation) + list of uploaded files — parses with `exceljs` instead of `xlsx`/SheetJS, which has two unpatched HIGH CVEs in its only published npm version
- `Experiments.tsx`: shows each graviscan experiment's linked waves inline; link a new wave / unlink an existing one (confirmation-gated) without a second accession picker
- Global upload-progress indicator in `Layout`, persists across navigation
- Fixes #286 (experiments now get an explicit `experiment_type`)
- Follow-up issue filed: #288 (`downloadImages` CSV export doesn't yet resolve wave-scoped metadata links)

## Process

- Proposal went through 3 rounds of the `openspec-review` skill (5 parallel adversarial subagents) before implementation, converging with no more real issues surfacing.
- Implemented with TDD, section by section, with a lint+typecheck+full-unit-suite gate before each new section.
- This branch was 5 commits behind `origin/main` by the time implementation finished; merged cleanly (no conflicts, including in `preload.ts`/`electron.d.ts` where both sides had additions).

## Test plan

- [x] `npm run lint` / `npm run format:check` — clean
- [x] `npm run test:unit` (full suite) — only pre-existing baseline failures remain (Windows path-separator assertions in `config-store`/`image-uploader`/`scan-coordinator`, one flaky `AccessionForm` timeout, one unrelated `MachineConfiguration` unhandled rejection) — none in files this change touches
- [x] `npx openspec validate add-graviscan-tier5-browse-metadata --strict` — valid
- [x] `npx electron-forge package` — succeeds, confirms the new routes/components compile into the packaged bundle
- [x] IPC coverage gate (`scripts/check-ipc-coverage.py`) — 95.6%, no regression
- [ ] **`npm run test:e2e`** — blocked in this dev environment by two pre-existing, unrelated infra issues documented in `tasks.md` Section 10: (1) every `*.e2e.ts`'s `launchElectronApp()` hardcodes `.webpack/main/index.js`, but Electron Forge's webpack plugin now nests output under `.webpack/x64/main/`; (2) `database.ts`'s `loadPrismaClient()` resolves the Prisma client via `process.resourcesPath`, which is always defined under Electron regardless of real packaging state, so a direct `electron .webpack/main/index.js` launch (as every e2e file does) can't find it. New E2E specs for this tier (`tests/e2e/graviscan-browse-metadata.e2e.ts`, `tests/e2e/graviscan-experiments-wave-linking.e2e.ts`) are written and `--list`-verified to parse, but unverified for pass/fail. These block the *entire* existing E2E suite, not just this tier's new specs — worth a separate fix.
- [ ] Manual golden-path walkthrough — not completed in this environment (no attached display; see `tasks.md` Section 12.6 for details)

See `openspec/changes/add-graviscan-tier5-browse-metadata/tasks.md` for the complete, itemized verification record.


eberrigan (commented):
> **Verdict: REQUEST_CHANGES (all fixed in 6ce0fc7)** — posted as comment, cannot request changes on own PR.

## Review Summary

Ran a 5-lens adversarial review (code quality/architecture, testing/TDD, scientific rigor/UX, security/cross-platform, behavioural correctness) against the actual PR diff, per this project's standard "cycle both ends" process. Security found nothing blocking. The other four lenses each independently surfaced real, confirmed bugs — several cross-verified by more than one lens — that the pre-implementation review and the unit-test suite's own mocks had both missed. All are fixed in 6ce0fc7.

1. **`ExperimentDetail.tsx` rendered raw `Date` objects as JSX children.** The real IPC payload for `capture_date`/`transplant_date` is an actual `Date` instance (structured-clone preserves it across `ipcRenderer.invoke`), not the `string` the type declared — React throws on a bare `Date` child. Masked because every test's fixture used ISO strings. Fixed with a `formatDate()` helper matching `BrowseScans.tsx`'s existing convention.
2. **False "success" message on failed metadata link.** `Experiments.tsx` showed "Metadata file successfully linked" even when the link call failed, because `useWaveMetadataLinks.link()` swallowed failures without telling the caller. `link()`/`unlink()` now return a boolean the callers check.
3. **Global upload-progress indicator was non-functional.** `Layout.tsx` read `progress.completed`/`progress.total`, but the real payload (`box-backup.ts`'s `BoxBackupProgress`) has no such fields — only `totalImages`/`completedImages`/`failedImages`/`currentExperiment`. Always showed "0/0". Fixed the type and rendering; corrected the test fixtures, which had been shaped to match the bug instead of the real payload.
4. **`BrowseGraviScans.tsx`'s per-row Box-backup progress never rendered.** Keyed by experiment id, but the real payload identifies the in-flight experiment by *name* — the lookup could never match. Fixed to key by name.
5. **"All Waves" mismatch warning (Decision 10) never fired**, even with diverged linked waves — the check only ever looked at the (nonexistent) selected wave when "All Waves" was chosen, directly contradicting this requirement's own "Mismatch warning before an 'All Waves' download" scenario. Fixed to check every linked wave and name every diverged one.
6. **No double-submit guard** on Link/Unlink in `ExperimentDetail.tsx` or the inline per-row Unlink in `Experiments.tsx` (unlike the attach panel's existing guard) — rapid re-clicks could race the backend's check-then-act pattern into a raw DB constraint error. Added pending-state guards to both.
7. **Stale-response race in `useWaveMetadataLinks`** — no cancellation guard meant an in-flight request for a previous `experimentId` could resolve after a newer one and overwrite state with the wrong experiment's data. Fixed with a ref-tracked "current experimentId" check before every `setState`.
8. **`unlink()` gave zero failure feedback**, unlike `link()`. Now sets `linkError` on failure, surfaced by both callers.
9. **E2E spec selector bug**: selected an experiment by an exact-string label that can never match the real rendered `"{species} - {name} ({scientist})"` text. Fixed to select by value/id.

## Important Issues (fixed)

- Task 5.1 in `tasks.md` claimed pagination-control test coverage that didn't actually exist. Added the real tests (Next/Previous → updated `offset`).

## Suggestions (deferred, tracked in tasks.md Section 13.11)

- `Experiments.tsx`'s attach panel fires a wasted no-op `listGraviMetadata('none')` call when no experiment is selected.
- `GraviMetadataList.tsx`'s Delete has no `window.confirm()`, inconsistent with Unlink's gating (backend still blocks deletion while referenced, so no actual data-loss risk).
- `setAuditLogger`'s unset-logger case fails silently rather than warning, if a graviscan mutation somehow ever occurs outside graviscan mode.
- `ensureDir`/`listScanFiles` are exposed in preload with no consumer in this diff yet (not a defect — the underlying main-process path-containment logic is sound and pre-existing — just unused surface for now).

## CI note

`Lint - Node.js` fails on this PR, but it also fails on `main`'s own latest commit (confirmed) — 6 pre-existing files with formatting drift unrelated to this change. Filed #291 to track; not fixed here to keep this diff scoped.

Full accounting of every fix, with file/line references, in `openspec/changes/add-graviscan-tier5-browse-metadata/tasks.md` Section 13.

---

eberrigan:
## Round 2: verifying the round-1 fixes (commit 68abce7)

Ran a second, narrower adversarial pass specifically against the round-1 fix commit (6ce0fc7) — checking whether those 9 fixes are actually correct/complete, and whether the fixes themselves introduced anything new.

**Verdict: no blocking issues.** All 9 fixes hold up:
- The `useWaveMetadataLinks` stale-response guard correctly rejects a cross-experiment stale response (verified by tracing mount→prop-change→late-resolve), and the ref-write-during-render pattern is React-Strict-Mode-safe.
- The double-submit guards, error surfacing, and mismatch-warning logic are all logically sound.

Two things worth closing before calling this fully converged, both addressed in 68abce7:
1. **Test-coverage gaps**, not code bugs: the "All Waves" mismatch-warning test never asserted the *matching* wave (0) was excluded from the warning — a "warn about every wave regardless of match" regression would've still passed it. Neither double-submit guard had a test actually exercising it. Added both.
2. **A narrow, real limitation surfaced (not introduced) by the `exp.name`-keying fix**: two *different* experiments sharing the same name, visible during a concurrent Box backup, would show bleeding progress on both rows — `Experiment.name` has no uniqueness constraint and the real payload has no id to key by instead. Filed #292 rather than fix in scope (would require changing `box-backup.ts`'s payload shape, which is pre-existing, untouched-by-this-PR code).

Full accounting in `tasks.md` Section 13 (13.12-13.13). Considering this converged — no further review rounds planned unless CI turns up something new.

eberrigan (commented):

## Review Summary

Five parallel adversarial reviews (Code Quality, Testing/TDD, Scientific Rigor & UX, Security & Cross-Platform, Behavioural Correctness) each independently read the current source rather than trusting `tasks.md`'s own account of the six bugs already found and fixed during this PR's CI investigation — all six are confirmed genuinely fixed. However, the reviews surfaced **recurrences of the exact same root-cause pattern that caused those six bugs** (an untyped/`any` IPC channel whose real envelope shape was never checked against the mock or the caller), landing in code this PR itself ships. Two of these are outright feature-breaking; two more are real data-integrity/UX gaps. Recommend fixing the blocking items below before merge — the pattern is well-understood by this point and the fixes should be small.

## Blocking Issues

1. **TIFF file preview is completely non-functional** — `src/renderer/ExperimentDetail.tsx`'s `FileRow.handleClick` reads `result.success`/`result.dataUri` directly off `readScanImage()`'s response, but `register-handlers.ts` wraps that channel in `wrapHandler`, which itself wraps `imageHandlers.readScanImage`'s own `{success, dataUri, error}` return in a second envelope — so the real wire shape is `{success: true, data: {success, dataUri, error}}`. `result.dataUri` is always `undefined`; the `` never renders. `readScanImage` is still typed `Promise<any>` in `electron.d.ts`, so TypeScript couldn't catch it, and `ExperimentDetail.test.tsx`'s mock uses the unwrapped shape (masking it, and never actually asserting an `` appears). This directly breaks the "inline TIFF preview" scenario in `ui-management-pages/spec.md`. Same root cause already fixed once this PR for `handleBackupToBox` — needs the same fix here.

2. **`handleBackupToBox` still misreports "upload already in progress" as success.** The envelope-unwrapping fix (bug 5) checks the outer `wrapHandler.success`, but never checks the *inner* domain result's own `success` field. `uploadAllScans`'s `uploadInProgress` guard resolves normally with `{success:false, uploaded:0, skipped:0, failed:0, errors:['Upload already in progress'], ...}` — none of `handleBackupToBox`'s branches match this, so it falls through to the generic success message and shows **"Uploaded 0 image(s), 0 skipped"** when the backup never actually ran. One line deeper than the already-fixed bug, same shape.

3. **A spreadsheet import can silently succeed with zero data written.** `guessMapping()` only auto-maps on an exact header match; if none match and the operator doesn't manually fill every required mapping, every row's required fields resolve to `''`. Row validation (`filled > 0 && filled < required.length`) only fires on *partial* fill, so an all-blank row raises no error. The import loop then skips every row (`if (!plateId) continue`), and the backend accepts an empty `plates` array with no length check, creating an orphaned `Accessions` row and returning `{success: true}`. The renderer never inspects `totalPlates` and unconditionally shows "Done uploading!" — a technician can be told the import succeeded while nothing landed in the database.

4. **Deleting a metadata file has no confirmation**, unlike every other destructive action added in this PR (Unlink in `ExperimentDetail.tsx`/`Experiments.tsx` both use `window.confirm`). `GraviMetadataList.tsx`'s `handleDelete` fires the IPC call immediately on click. The backend does block deletion while a file is still referenced, so this can't silently corrupt in-use data — but for an unreferenced file, one accidental click permanently cascades away all its plate/section rows with zero warning. Flagged independently by two reviewers; already named as "deferred" in `tasks.md` 13.11 but still unfixed.

5. **No boundary guard or double-submit guard on two other new flows:**
   - `BrowseGraviScans.tsx`'s pagination: `graviscansBrowseByExperiment`'s `total` field is fetched but discarded, and the `Next` button has no disabled condition — clicking past the last page renders "No GraviScan data is present," indistinguishable from a genuinely empty result.
   - `GraviMetadataUpload.tsx`'s `Import` button has no `isImporting`/disabled guard (every comparable action elsewhere in this PR has one), and the backend has no uniqueness check on the accession name — a rapid double-click creates duplicate metadata-file records from one spreadsheet.

## Important Issues

- **Box backup will silently fail on most real lab Windows machines, unverified by CI.** `box-backup.ts`'s `rcloneCopyFiles()` calls `fs.symlinkSync` directly instead of the codebase's own `ensureSymlinkOrCopy()` (whose doc comment exists specifically because unprivileged Windows accounts can't create symlinks). Since rclone itself isn't installed on any CI runner, this code path has never executed anywhere in CI — the exact "never exercised, hidden" pattern this PR's own investigation already found six times. Caught by the outer try/catch (no crash), but Box backup would be fully non-functional on a stock Windows lab machine with rclone installed.
- **CSV/formula injection risk in the Box metadata export.** `exportMetadataCSV()`/`csvEscape()` only escape commas/quotes/newlines; operator-entered `custom_note`/`accession`/`plate_barcode` values starting with `=`, `+`, `-`, or `@` flow unescaped into a CSV that's uploaded to Box for humans to open in Excel — classic formula-injection surface in a shared-lab context.
- **The "Filename" column in `ExperimentDetail.tsx` shows `scan.id` (a UUID), not an actual filename** — there's no derived human-readable name, so a scientist can't visually match a table row to the real TIFF on disk or in Box.
- **`onUploadProgress`'s payload is still `Promise<any>`/untyped `unknown`,** with three independently hand-typed, unlinked mirrors of the same shape (`BoxBackupProgress` in main, `UploadProgress` in `BrowseGraviScans.tsx`, `UploadProgressData` in `Layout.tsx`, the last consumed via an unchecked `as` cast). This is exactly the pattern already fixed once for `uploadAllScans` in this same PR — not generalized to the sibling channel.
- **Same-experiment wave link/unlink race across two independent UI surfaces.** `WaveMetadataLinksContext`'s `link()` triggers a full `refetch()` while `unlink()` does a local optimistic filter; each is guarded only by its own caller's local pending flag (attach panel vs. the row's own Unlink button are separate components with separate flags). A link from one surface and an unlink from the other, fired concurrently for the same experiment, can have a late `refetch()` silently revert the other's just-applied change. Same bug class as the already-fixed cross-experiment staleness bug, one layer narrower (same-experiment, cross-surface).
- **Global upload-status banner never auto-clears** after an upload finishes (no "complete" signal, just the last progress tick), and its dismiss check uses reference equality, so any subsequent duplicate/stale progress tick silently un-dismisses it.
- **No warning when navigating away mid-spreadsheet-upload** in `Metadata.tsx` — if manual column-mapping was required (see Blocking #3), that work is silently discarded on any sidebar click.
- **Stale test mocks recur in `App.test.tsx` and `UploadStatusContext.test.tsx`** — both still use the pre-fix, unwrapped/fake shapes for `uploadAllScans`/progress payloads that `BrowseGraviScans.test.tsx` and `Layout.test.tsx` were correctly updated to use. Currently harmless (these files don't exercise those branches) but the same landmine pattern, left in place a third time.
- **The `useWaveMetadataLinks` "stale-response race" regression test doesn't test the race it claims to** — it only resolves two *different* `experimentId` keys, which trivially can't clobber each other in a map keyed by id; the actual at-risk same-key race (see above) has no test.
- Box-backup's partial-failure message shows only the first error and never states how many images *did* succeed, so "9/10 succeeded" and "0/10, all failed" read almost identically.
- `handleBackupToBox` has a `finally` but no `catch` — if the IPC call itself rejects (not just resolves with `success:false`), the button silently resets with zero user-facing message.
- The original feature implementation (Sections 1-9) lands as a single squashed commit, so test-first discipline for the *initial* build isn't independently verifiable via git history — only the later bug-fix/review-response phases show granular red→green commits.

## Suggestions

- No server-side size cap on `graviscan:parse-excel-file`'s buffer (the 15MB check is renderer-only) — cheap defense-in-depth now that parsing was deliberately moved to the main process.
- `resolveGraviScanPath()` isn't routed through the same path-containment check as renderer-facing handlers — low risk today (DB-authored, not renderer-authored paths) but worth a comment documenting the trust boundary.
- Symlink basenames in `rcloneCopyFiles`'s shared temp dir could collide across waves with identical filenames — no uniqueness prefix.
- `useResizableColumns` can get stuck mid-drag if `mouseup` fires outside the OS window bounds (multi-monitor/non-maximized) — not a crash, but a real stuck-state until the next drag starts.
- `fetchExperiments` has no stale-response guard — `updateFilter` can trigger two concurrent requests (an effect-driven one plus its own explicit debounced/immediate call) with no abort/request-id check.
- Several "TEMPORARY DIAGNOSTIC — remove once resolved" comments remain in `tests/e2e/*.e2e.ts` files even though the investigations they reference are now resolved per `tasks.md` — either resolve task 14.9's keep/trim decision before merge or update the comments so they don't read as stale.
- Magic numbers (300ms filter debounce, 1500ms post-import reset) aren't extracted as named constants, inconsistent with `PAGE_SIZE`/`MAX_FILE_SIZE` a few lines away in the same files.
- `metadataLinkingAvailable` is computed in the main process but never surfaced in the renderer.
- `ExperimentDetail.tsx` sizes a `wave` column via `useResizableColumns` but never renders its resize handle — dead state.
- `excel-parser.ts`'s row iteration silently skips blank rows, so a validation error's reported "Row N" (an array index) can drift from the true spreadsheet row.

---

Issue #299, opened 2026-08-05
CylinderScan: Basler camera acquisition-metadata readback (confirm actual settings used, not just requested)


eberrigan:
## Problem

There is no readback capability for actual Basler camera acquisition attributes anywhere in `python/hardware/camera.py` or `scanner.py` (confirmed by reading both files during the accompanying CylinderScan delete/upload-integrity change, 2026-08-05). `metadata.json` records the settings *requested* by the user/UI, not attributes confirmed from the camera after acquisition — so if the camera silently clamps, rejects, or defaults a requested value, that divergence is invisible.

This mirrors [`eberrigan/bloom-desktop-pilot#3`](https://github.com/eberrigan/bloom-desktop-pilot/issues/3): "We should save all possible attributes from the Basler API so camera settings per scan are known."

## Why this needs its own scoped design, not a quick patch

This isn't a small addition — it has a real metadata-timing tradeoff to resolve first:

- Current behavior writes `metadata.json` **before** capture starts, specifically so metadata survives partial/failed captures.
- Acquisition readback (the actual, camera-confirmed values) is only available **after** acquisition starts/completes.
- Reconciling these needs either a second write pass (with its own partial-failure semantics to define) or an explicit, documented decision to accept the tradeoff of readback-vs-pre-capture-durability.

This also needs new PyPylon instrumentation work in the Python layer, not just a TypeScript/Electron change.

## Scope for whoever picks this up

- Design the write-timing approach (second write pass vs. accepted tradeoff) before implementing.
- Add PyPylon readback calls for the relevant acquisition attributes in `python/hardware/camera.py`.
- Decide which attributes are worth confirming (exposure, gain, gamma, frame rate, etc. — camera settings actually in use today) vs. pilot #3's broader "all possible attributes" framing.

Issue #298, opened 2026-08-05
CylinderScan: scheduled upload/storage audit tool, with reconciliation/write-back for historically-corrupted uploads


eberrigan:
## Problem

There is no automated audit mechanism to detect drift between local `Image.status` and actual Supabase storage/database state, mirroring [`eberrigan/bloom-desktop-pilot#61`](https://github.com/eberrigan/bloom-desktop-pilot/issues/61).

As of the accompanying CylinderScan delete/upload-integrity change (2026-08-05), freshly-uploaded images are independently verified against Supabase storage before being marked `'uploaded'` locally, and retry now skips already-`'uploaded'` images entirely rather than re-uploading them. This closes the *forward-going* version of the problem, but leaves a real, explicitly-acknowledged gap:

**Historically-corrupted data is unreachable by anything in that tier.** An image that was already wrongly marked `'uploaded'` by the pre-existing bug (before the new verification existed) — including any scan uploaded before that change ships — has no path back to a re-uploadable state. Neither `BrowseScans.tsx` nor `ScanPreview.tsx` offers a "verify"/"re-check" affordance separate from Upload/Retry, and both disable Upload/Retry once a scan's images are all locally `'uploaded'` — exactly the state the historical bug produces.

## Scope (explicitly widened beyond pilot #61's original "detect and report")

This tool needs to do more than report drift:

1. **Detection**: walk local `Image` rows marked `'uploaded'` and confirm each still has a corresponding object in Supabase storage (and ideally a matching `cyl_images` row).
2. **Reconciliation / write-back**: flip a confirmed-corrupted `'uploaded'` image back to a re-uploadable local status once drift is confirmed, so retry can pick it up. This is genuinely new scope, not just detection — a passive report is not enough since there's currently no UI path to act on a report even if one existed.
3. **Lookup strategy for historical data**: there is no local↔cloud join key (see the companion UUID-traceability issue, filed alongside this one), so this tool needs its own strategy for matching local rows to cloud objects for scans that predate any future traceability fix — e.g. a business-key join on `(plant_qr_code, frame_number, date_scanned)`, or a local schema addition scoped to this tool's own design. Whichever local-UUID/traceability fix eventually lands (tracked separately) may simplify this for new data going forward, but this tool still needs to handle the historical backlog either way.

## Also folds in: automatic in-line recovery (considered and deferred, not this tool by default)

[...]

Issue #297, opened 2026-08-05
CylinderScan: no local↔cloud UUID traceability; delete-then-rescan can orphan indistinguishable duplicate cloud data


eberrigan:
## Problem

No local scan/image identifier is ever sent to the cloud row when uploading CylinderScan images. `cyl_images` (per the generated `database.types.d.ts`) has columns `id, scan_id, frame_number, date_scanned, object_path, status, uploaded_at` — no slot for a local id — and the `CylImageMetadata` type `@salk-hpi/bloom-fs`'s `uploadImages()` accepts has no id field either. There is no way to trace a cloud record back to the local DB row that produced it.

This mirrors [`eberrigan/bloom-desktop-pilot#59`](https://github.com/eberrigan/bloom-desktop-pilot/issues/59), which already identifies the same gap and proposes adding `local_image_id`/`local_scan_id` columns.

Fixing this requires changes outside this repo's reach:
- A `@salk-hpi/bloom-fs` package type change (to accept/pass a local id)
- A Supabase schema migration (`cyl_images` needs a column for it, and the `insert_image_v3_0` RPC needs to accept it)

## Related prior art

[`eberrigan/bloom-desktop-pilot#60`'s 2026-06-17 follow-up comment](https://github.com/eberrigan/bloom-desktop-pilot/issues/60#issuecomment-4725623888) proposes a structural fix that solves this directly: make the storage path a pure function of `(local_scan_uuid, frame_number)` (`images/cyl-images/<local_scan_uuid>/<frame_number>.png`), so the local UUID *is* the path — trivial cross-system traceability, no schema migration required for the path itself (an optional `sha256` column is discussed as a follow-on for integrity, not required for traceability). That comment is worth reading in full before scoping a fix here; it also has a smaller-effort recommendation ("(A) now, (B) when there's bandwidth") that's directly usable as this issue's implementation plan.

## Second, related gap found during CylinderScan delete/upload hardening (2026-08-05)

While implementing scan deletion and upload verification (see the accompanying change), we found this traceability gap has a second consequence beyond "can't join local↔cloud data": **delete-then-rescan can leave two independently-valid, fully-verified image sets in cloud storage with no way to tell which is authoritative.**

- Deleting a CylinderScan is soft-delete-only and purely local — it never touches Supabase.
[...]

Issue #110, opened 2026-03-04
Evaluate upload concurrency: pilot uses 10 workers vs bloom's 4


eberrigan:
## Summary

The pilot uses `pLimit(10)` for concurrent image uploads (10 simultaneous uploads), while bloom-desktop uses 4 concurrent upload workers. The optimal concurrency depends on network bandwidth, Supabase rate limits, and image sizes. This should be evaluated and potentially made configurable.

## Pilot behavior

Upload uses `pLimit(10)` — up to 10 simultaneous image uploads to Supabase Storage.

## Current bloom-desktop status

ImageUploader uses 4 concurrent upload workers.

## Acceptance criteria

- [ ] Benchmark upload performance with different concurrency levels (4, 8, 10)
- [ ] Consider making upload concurrency configurable
- [ ] Document the chosen concurrency and rationale

## References

- **Pilot files:** `src/main/upload.ts`
- **Bloom-desktop files:** `src/main/image-uploader.ts`
- Priority: P2 (nice-to-have / UX polish)

_From feature parity analysis (bloom-desktop-pilot vs bloom-desktop)_

eberrigan:
Update from the CylinderScan delete/upload-integrity hardening (2026-08-05): this change added a documentation comment to `image-uploader.ts`'s `nWorkers: 4` constant, matching GraviScan's existing rationale (3 round-trips per image: insert RPC → file upload → update RPC) — no behavior change. That satisfies this issue's third acceptance criterion ("Document the chosen concurrency and rationale").

The other two acceptance criteria are explicitly **not** addressed by that change and remain open here:
- [ ] Benchmark upload performance with different concurrency levels (4, 8, 10)
- [ ] Consider making upload concurrency configurable

Leaving this issue open for those two.

PR #304, opened 2026-08-05
fix(cylinderscan): delete/upload data-integrity hardening + duplicate-scan check (Tier 2)

### LINKED ISSUE #120: Block duplicate scans with same plant/wave/age in UI
## Problem

The UI allows users to scan the same plant with the same wave number and plant age multiple times across different dates. The current duplicate check in [`CaptureScan.tsx:88-133`](src/renderer/CaptureScan.tsx#L88-L133) only checks if the **same plant was scanned today** in the same experiment. It does not check the `wave_number + plant_age_days + plant_id` combination, which should be unique.

### Current behavior
- Plant A, Wave 2, Age 14 scanned on March 1 → allowed ✅
- Plant A, Wave 2, Age 14 scanned on March 5 → allowed ❌ (should be blocked)

### Expected behavior
- The scan button should be **disabled** (blocked, not just a warning) when a scan already exists with the same `plant_id + wave_number + plant_age_days` combination.

## Root Cause

The `checkDuplicateScan` effect only queries `getMostRecentScanDate` (plant + experiment, same day). There is no query or IPC handler that checks for the `plant_id + wave_number + plant_age_days` combination.

### 1. New IPC handler in [`database-handlers.ts`](src/main/database-handlers.ts)

Add a new handler `db:scans:checkDuplicate` that queries:
```typescript
const existing = await db.scan.findFirst({
  where: {
    plant_id: plantId,
    wave_number: waveNumber,
    plant_age_days: plantAgeDays,
    deleted: false,
  },
  select: { id: true, capture_date: true },
});
```
This uses the existing `plant_id` index. No schema changes needed.

### 2. Update preload/IPC types

Expose the new handler through the preload bridge so the renderer can call it.

### 3. Update [`CaptureScan.tsx`](src/renderer/CaptureScan.tsx)

Replace or augment the existing `checkDuplicateScan` effect (lines 88-133) to call the new handler with `plant_id + wave_number + plant_age_days`. The existing `canStartScan` logic (line 379-384) already disables the Start Scan button when `duplicateScanWarning` is set, so the blocking behavior is already wired — just needs the right check.

Update the warning message to be specific, e.g.:
> "A scan already exists for this plant with wave 2 and age 14 days (scanned on 2026-03-01)"

### 4. Remove the same-day-only check

The current `getMostRecentScanDate` check becomes redundant once the new check is in place. Remove it or replace it entirely.

### Unit tests
- [ ] `database-handlers.test.ts`: Test `db:scans:checkDuplicate` returns match when duplicate exists
- [ ] `database-handlers.test.ts`: Test returns null when no duplicate exists
- [ ] `database-handlers.test.ts`: Test ignores soft-deleted scans (`deleted: true`)

### E2E tests
- [ ] Fill metadata form with plant/wave/age that already exists in DB → Start Scan button is disabled, warning shown
- [ ] Fill metadata form with unique plant/wave/age → Start Scan button is enabled, no warning
- [ ] Change wave number on a duplicate → warning clears, button re-enables
- [ ] Change plant ID on a duplicate → warning clears, button re-enables

### Component tests (Vitest + React Testing Library)
- [ ] `CaptureScan` renders warning when duplicate check returns a match
- [ ] `CaptureScan` hides warning when duplicate check returns no match
- [ ] Start Scan button disabled state matches duplicate warning state

eberrigan:
## Summary

CylinderScan finalization roadmap, Tier 2: closes two confirmed live upload data-integrity bugs (retry duplicate-row risk, unverified upload success), finishes the soft-delete feature (metadata.json sync + ScanPreview delete affordance), and replaces an imprecise duplicate-scan check with the correct 4-field key.

## Changes

- **Delete completion (#79, #105)**: `db:scans:delete` now also writes `deleted: true` into the scan's on-disk `metadata.json`. Added a delete affordance to `ScanPreview.tsx`. Both it and `BrowseScans.tsx` now share a real `DeleteConfirmModal` (Plant ID + capture date) instead of `BrowseScans`' previous generic `window.confirm()`. `BrowseScans.tsx`'s Delete button is now also disabled while that scan's upload is in flight.
- **Upload data-integrity fixes (`image-uploader.ts`)**:
  - `uploadScan`/`uploadBatch` now skip soft-deleted scans entirely.
  - Retry now skips images already marked `'uploaded'` instead of resending them — closes a duplicate-remote-row risk. A regression test guards the filtered-array index-safety this depends on (a naive `scan.images[index]` implementation would silently apply status updates to the wrong image).
  - Freshly-uploaded images are now independently verified against Supabase storage (not just trusted from bloom-fs's callback) before being marked `'uploaded'` locally, with a bounded 3-attempt retry (500ms delay) distinguishing confirmed-missing from a transient lookup failure. This closes a real incident the project owner described directly: a failed-then-retried upload was marked succeeded without the bytes actually landing in storage.
  - `uploadScan()` now explicitly awaits verification work (bounded at the same `nWorkers` concurrency as upload, via bloom-fs's own `concurrentMap`) so it can't return before trailing images' status writes land.
- **Duplicate-scan check (#120)**: new `db:scans:checkDuplicate` handler keyed on `(plant_id, experiment_id, wave_number, plant_age_days)`, replacing `CaptureScan.tsx`'s previous same-day/(plant_id + experiment_id) check — the old key missed `wave_number`/`plant_age_days` and could false-positive across experiments. `db:scans:getMostRecentScanDate` removed as dead code (preload, types, and all its test coverage).
- **Test-infrastructure fixes found along the way**: a new vitest `globalSetup` builds a pristine template SQLite DB once before any test file starts, fixing an FK-violation crash that occurred when an isolated test file copied the *live*, actively-mutated `dev.db` mid-suite. A race in `capture-scan-config.test.tsx` between session-restore and test input was also fixed.

**BREAKING**: `db:scans:getMostRecentScanDate` is removed.

## OpenSpec Change

- Change ID: `add-cylinderscan-delete-upload-integrity`
- Affected capabilities: `ui-management-pages`, `upload`
- Delta types: MODIFIED (both)
- All `tasks.md` items complete: yes (Section 5's follow-up-issue tasks and Section 6's E2E task are marked complete/deferred with reasoning inline, not silently skipped)

### TypeScript Tests

- [x] Unit tests pass (`npm run test:unit`) — 1379 passed, 9 skipped, 7 pre-existing failures across 4 files, all path-separator/timing issues confirmed present on the pre-this-PR baseline via `git stash` diffing (unrelated to this change). Verified stable across 3+ full-suite runs.
- [x] Unit tests added for new functionality — `scan-metadata-json.test.ts`, `scans-delete.test.ts`, `DeleteConfirmModal.test.tsx`, `BrowseScans.test.tsx`, `ScanPreview.test.tsx`, `image-uploader.test.ts` (12 new tests), `scans-check-duplicate.test.ts` (15 tests), `capture-scan-config.test.tsx` (6 new tests)
- [ ] Coverage meets 50% threshold — not separately re-checked this run; `npm run lint`/`tsc`/unit suite all green

### Python Tests

- [ ] N/A — no Python changes in this PR

### Integration Tests

- [ ] IPC tests — covered via the E2E IPC coverage gate below, not the separate `test:ipc` script
- [x] Database/IPC coverage gate (`scripts/check-ipc-coverage.py`) — 95.6% (43/45 handlers), `db:scans:checkDuplicate` has 2 real E2E test calls; the 2 untested handlers predate this PR

### E2E Tests

- [ ] E2E tests (`npm run test:e2e`) — **deferred to CI.** This repo's Electron E2E tests have no true headless mode on Windows (CI's own headless run depends on Linux + Xvfb, which isn't available locally here); two local attempts each popped up unexpected visible windows on the developer's screen (once from a stale `BLOOM_DATABASE_URL` override, once from this worktree never having been built), and the user chose to let CI's run be the real verification rather than attempt a third local run. The new/rewritten E2E tests (`renderer-database-ipc.e2e.ts`'s `checkDuplicate` coverage, `plant-barcode-validation.e2e.ts`'s rewritten duplicate-scan scenarios) were reviewed by hand for correctness against the new 4-field key and the direct-Prisma-seed pattern this file already uses.

### Hardware Integration

- [ ] N/A — no hardware-facing changes

## Build & Packaging

- [x] TypeScript compiles without errors (`npx tsc --noEmit`)
- [ ] Packaged app / `npm run package` — not run this PR; no build-config or native-module changes

## Linting & Formatting

- [x] Prettier formatting applied (`npm run format:check`) — clean for every file this PR touches; a pre-existing prettier non-idempotency bug in `tasks.md` (missing blank lines before nested markdown lists, causing indentation to flip-flop forever under repeated `--write`) was found and fixed as part of getting this gate green
- [!] 6 files elsewhere in the repo (`docs/superpowers/plans/2026-08-03-cylinderscan-finalization-roadmap.md`, 5 files under `openspec/specs/`) still fail `format:check` — pre-existing on `main`, untouched by this PR

## Database Changes

- [x] No Prisma schema changes — this tier deliberately stays within the existing schema (see design.md Decision 9's rejected alternative)

- [ ] CI tests pass on Linux, macOS, and Windows — pending this PR's CI run
- [x] File paths use `path.join()`/`path.isAbsolute()` consistently with the existing `image-uploader.ts` pattern

## Breaking Changes

- [x] Breaking changes documented below

`db:scans:getMostRecentScanDate` (IPC handler, preload exposure, and `electron.d.ts` type) is removed. Confirmed via full-tree grep that `CaptureScan.tsx` (its only caller) was migrated to `db:scans:checkDuplicate` in the same PR, so there is no remaining caller.

## Related Issues

Closes #120 (duplicate-scan check by plant/wave/age)
Progresses #79, #105 (scan delete — comment posted explaining the three acceptance criteria intentionally not met by this soft-delete-only implementation)
Related: #297 (local↔cloud UUID traceability + delete-then-rescan cloud-storage orphaning, filed as follow-up), #298 (widened upload audit/reconciliation tool, filed as follow-up), #299 (Basler acquisition-metadata readback, filed as follow-up), #110 (upload worker count — commented with the now-satisfied rationale criterion, two asks left open)

## Reviewer Notes

- The upload-verification and retry-skip logic in `image-uploader.ts` is the highest-stakes part of this PR — it's a direct response to a data-integrity incident described firsthand by the project owner (a retried upload that erred, then silently succeeded-looking without the image actually present in storage). `design.md` Decisions 7-10 have the full rationale for what's in scope vs. explicitly deferred (e.g., no re-verification of already-`'uploaded'` images from *before* this PR ships — that's the widened #298 follow-up's job).
- `openspec/changes/add-cylinderscan-delete-upload-integrity/` went through 3 rounds of adversarial multi-agent review before implementation started; `design.md` documents two real mid-review redesigns (Decision 9's rescoping, Decision 8's await fix) if you want that history.
- Please weigh in on whether deferring `npm run test:e2e` to CI (rather than running it locally first) is acceptable here, given the local-environment constraints described above.


eberrigan (commented):

## Review Summary

Five independent subagents reviewed this PR adversarially (code quality/architecture, testing/TDD discipline, scientific rigor/UX, security/cross-platform, behavioural correctness/edge cases), each reading the actual source and test files rather than trusting the PR description. Two of the five (code quality and behavioural correctness) independently converged on the same real bug in the upload-verification path, which has now been fixed and covered by a regression test. All other findings were either already-adequately-mitigated (verified empirically, not just asserted), pre-existing and out of scope, or addressed as documentation/copy improvements. No blocking issues remain open.

## Fixed as a result of this review

- **The per-entry `catch` block in `uploadScan()`'s verification pass never wrote a terminal DB status.** If verification threw unexpectedly (e.g. a Supabase call escaping the normal `{data,error}` tuple pattern) rather than resolving to a handled outcome, the image was left at local status `'uploading'` forever — and the next retry's `status !== 'uploaded'` filter would re-upload it, reintroducing the exact duplicate-remote-row bug this tier exists to close. Fixed in `src/main/image-uploader.ts` by writing `status:'failed'` in that branch too, matching its two sibling branches. New regression test in `tests/unit/image-uploader.test.ts` simulates the unexpected-throw case, fails against the old code, passes against the fix.
- `scansDelete`'s metadata-sync failure log no longer conflates the expected "legacy scan predates metadata.json" case (`ENOENT`) with a real unexpected failure (corrupt JSON, permission denied) — the latter now logs at `error` level with a distinguishing message. New tests in `tests/unit/scans-delete.test.ts` cover both branches.
- `docs/DATABASE.md` still documented the removed `getMostRecentScanDate` handler as live; updated to document `checkDuplicate`.
- `DeleteConfirmModal` now tells the user up front that already-uploaded images are not removed from cloud storage — surfaces design.md Decision 14's delete-then-rescan orphaning risk at the moment of the destructive click, not just in a design doc. New test added.
- `design.md`'s Risks section now documents that verification runs only after a whole upload batch resolves, not per-image as each completes — widening the crash-during-upload window versus a per-image approach. This is a deliberate, now-documented trade-off (a per-image fix would reintroduce the unbounded-concurrency problem Decision 8 exists to prevent), cross-referenced to the #298 follow-up, which already needs to detect and repair this same class of gap. Commented on #298 with the specific scenario.

## Investigated, no change needed

- A suspected double-click race on the delete-confirmation modal's Confirm button (no `disabled` state on the button). Wrote a test that fires two rapid clicks and asserts `scans.delete` is called only once — it **passed against the unmodified code**, because `setScanPendingDelete(null)`/`setShowDeleteConfirm(false)` run synchronously before any `await`, and React's synchronous batched update unmounts the modal before a second click can register, both in the test environment and under React 18's real batching. Removed the test rather than keep a defensive fix for a non-reproducible risk.

## Noted, out of scope for this PR

- `db:scans:upload` has no in-flight-per-scan concurrency guard (two concurrent calls for the same scan would both upload every image) — confirmed pre-existing on `main`, not introduced by this PR.
- `db:scans:create` accepts an unvalidated `path` field over IPC — confirmed pre-existing, and confirmed unreachable with attacker-controlled data via any current renderer call site (the only real creation path sanitizes `plant_id` before it ever becomes `scan.path`).
- `uploadScan()` has grown to ~250 lines across five responsibilities; a good candidate for extracting the verify-and-write closure into its own method in a future cleanup, not blocking.
- `VERIFICATION_CONCURRENCY` and the pre-existing upload `nWorkers` are two separately-hardcoded `4`s that happen to match today, not derived from one source — could silently drift if #110's concurrency-tuning follow-up changes one without the other.
- The 500ms verification retry delay is exercised by fake-timer tests but not independently pinned to exactly 500ms.

## CI

All 18 checks green, including all three platforms' full E2E Dev Build jobs (Windows, macOS, Ubuntu) and the E2E IPC coverage gate (95.6%, `checkDuplicate` covered).

---
*Review by Claude Code subagent team (Code Quality · Testing/TDD · Scientific Rigor & UX · Security & Cross-Platform · Behavioural Correctness). GitHub Copilot review is not configured for this repository, so no Copilot comments were available to factor in.*

Issue #303, opened 2026-08-05
box-backup.ts's rcloneCopyFiles() has the same unguarded-symlink Windows bug as the Prisma client fix, and is untested


eberrigan:
## Context

While implementing `add-cylinderscan-packaging-ci` (Tier 5a of the CylinderScan finalization roadmap), the new packaged-app CI verification caught a real, pre-existing Windows bug: `src/main/database.ts` tried `fs.symlinkSync()` to link the Prisma client outside the asar archive, which fails with `EPERM` on Windows without Developer Mode/admin privileges (the default state for most accounts). That was fixed via a new `ensureSymlinkOrCopy()` helper (`src/main/fs-symlink-or-copy.ts`) that falls back to a real recursive copy when the symlink attempt fails.

## The same bug class exists in `box-backup.ts`, unfixed

`src/main/box-backup.ts:88` — inside `rcloneCopyFiles()` — calls `fs.symlinkSync(resolvedPath, linkPath)` unconditionally, with **no platform guard**, to stage a specific file subset into a temp directory before backing raw TIFFs up to Box via `rclone copy`. It's invoked from the normal upload flow (`src/main/graviscan/image-handlers.ts:219-221`) on every platform, including packaged Windows builds, whenever `rclone` is installed (plausible on a lab machine, since Box backup is the whole point of the feature).

This has a different risk profile than the Prisma case, which is why it wasn't fixed as part of the same change:

- The Prisma symlink is a **one-time**, small (few-MB) link created once at app startup.
- `box-backup.ts`'s symlinks stage **scan images** (potentially large TIFFs) **repeatedly**, on every upload batch.

Blindly reusing `ensureSymlinkOrCopy()`'s copy-fallback here could mean copying large files on every batch upload on any Windows machine without symlink privilege — a real disk/perf regression, not obviously the right fix. This needs its own design decision, e.g.:

- Skip the symlink-staging approach on Windows entirely and pass `rclone` a `--files-from`-style list of the real file paths directly, avoiding the temp-directory-of-symlinks pattern altogether.
- Or apply the same copy-fallback but only for Windows and log a warning about the disk-space cost.

## Also currently untested

[...]

Issue #296, opened 2026-08-05
Define a release cycle for Windows (CylinderScan) and Linux (GraviScan) builds


eberrigan:
## Context

`add-cylinderscan-packaging-ci` (CylinderScan Tier 5a) and its GraviScan counterpart (#294/#295) add CI verification that `npm run make` produces working installer artifacts, and upload those artifacts as transient GitHub Actions build artifacts (7-30 day retention, tied to a single CI run) for manual pickup. That is deliberately *not* a release process — there's no versioning scheme, no signing/notarization, no persistent publish target, and no cadence.

This repo has no release workflow today (confirmed: `.github/workflows/` contains only `pr-checks.yml`).

## Ask

Scope an actual release pipeline for both deployment targets — Windows lab machines (CylinderScan) and Linux lab machines (GraviScan):

- **Versioning**: where does the version number live (`package.json`? git tags?), how does it bump, is there a changelog process (there's an existing `update-changelog` workflow/command — check whether it already covers this or needs extending).
- **Signing/notarization**: `forge.config.ts`'s makers (`MakerSquirrel`, `MakerDMG`, `MakerDeb`/`MakerRpm`) are currently unsigned stubs. Decide whether lab-machine deployment requires signing (macOS notarization/Gatekeeper, Windows Authenticode) or whether unsigned installers are acceptable for this deployment model (internal lab machines, not public distribution).
- **Publish target**: GitHub Releases is the obvious default (free, versioned, durable) — confirm or pick an alternative.
- **Cadence/trigger**: tag-push-triggered release build, manual workflow_dispatch, or something else.
- **Scope**: does this need to happen for both platforms simultaneously, or can Windows/CylinderScan ship first (it's further along per the finalization roadmap) with Linux/GraviScan following once its own packaging-CI tier (#294/#295) lands?

## Related

- `add-cylinderscan-packaging-ci` (adds CI build-artifact retention, not a release process)
- #294, #295 (GraviScan Linux packaging CI + deployment docs — this release-cycle work likely depends on that landing first for the Linux side)
- Tier 5b of the CylinderScan finalization roadmap (`docs/superpowers/plans/2026-08-03-cylinderscan-finalization-roadmap.md`) — manual Windows install/QA on real hardware; a real release pipeline would supersede/formalize part of that manual process over time

eberrigan:
[...]

Issue #295, opened 2026-08-05
[GraviScan] Document Linux packaged-app deployment (libusb-filter/LD_PRELOAD, deb vs rpm, systemd/permissions)


eberrigan:
## Context

Confirmed via research while scoping `add-cylinderscan-packaging-ci` (the CylinderScan/Windows+macOS packaging-CI change): GraviScan's Linux *driver* setup is documented in depth (`docs/GRAVISCAN_SCANNER_DRIVER_SETUP.md` covers the SANE/Epson `iscan`/`epkowa` backend install, udev rules, etc.), but the *packaged-app* deployment side has no documentation anywhere in the repo:

- `forge.config.ts`'s Linux-only `extraResource` bundles `./src/main/native/libusb-filter.so`, described only as "LD_PRELOAD USB filter for parallel scanner isolation (Linux only)," built by `npm run build:native` (`scripts/build-libusb-filter.sh`). No doc explains how/when `LD_PRELOAD` actually gets wired in at runtime for a packaged install, or what breaks if it isn't.
- No doc explains why `MakerDeb`+`MakerRpm` were chosen (and not e.g. AppImage), or which the actual lab machines need.
- `docs/PACKAGING.md`'s only Linux content is a one-line troubleshooting note ("Error loading shared libraries" → install `openssl`/`libssl3`) — no systemd, permissions, or udev-interaction-with-a-packaged-install content.

## Ask

Write deployment documentation (a new doc, or a new section in `docs/PACKAGING.md`/`docs/GRAVISCAN_SCANNER_DRIVER_SETUP.md`) covering:

- How/when the `libusb-filter.so` `LD_PRELOAD` mechanism is applied for a packaged (not dev-mode) install, and what the failure mode looks like if it isn't.
- The deb-vs-rpm (vs. AppImage) choice, and which the real lab machines actually need.
- Any systemd/permissions caveats specific to running the packaged app (not just the SANE driver) on a GraviScan lab machine.

## Related

- #294 (companion issue: CI verification for the Linux packaging this documents)
- Proposed as a new tier ("Packaging CI & Linux Deployment Docs") in `docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md`

eberrigan:
[...]

Issue #294, opened 2026-08-05
[GraviScan] Add Linux packaging CI job (npm run make + artifact retention)


eberrigan:
## Context

Tier 5a of the CylinderScan finalization roadmap (`add-cylinderscan-packaging-ci`, closing #251/#57) added `npm run make` CI verification (maker-artifact existence, window-render assertion, DB-init check) for macOS and Windows, plus uploading the resulting DMG/Squirrel artifacts as downloadable CI artifacts. GraviScan's deployment target is Linux, which was explicitly out of scope for that change (deferred by design, not an oversight).

Confirmed via research while scoping that change: no CI job currently packages, verifies, or uploads Linux artifacts (`.deb`/`.rpm`/AppImage) at all. `forge.config.ts` already configures `MakerDeb`/`MakerRpm` (no AppImage maker configured today), but `npm run make` is never exercised in CI on any platform, and no packaged-app launch check exists for Linux.

## Ask

Mirroring the macOS/Windows `test-make`/`test-make-windows` jobs added in `add-cylinderscan-packaging-ci`:

- Add a `test-make-linux` CI job: run `npm run make` on `ubuntu-latest`, verify the deb/rpm maker artifacts exist and are non-trivially sized, launch the staged app binary (likely needs `xvfb-run`, unlike macOS/Windows which have a real desktop session) and assert a window renders + DB initializes.
- Upload the resulting `.deb`/`.rpm` artifacts as downloadable CI artifacts (matching the macOS/Windows artifact-upload pattern added in the same change).
- Consider whether an AppImage maker should be added given it's a common Linux distribution format for machines that aren't running a specific package manager — the lab's actual GraviScan machines' OS/package-manager should inform this, not left as an assumption.

## Related

- `add-cylinderscan-packaging-ci` (CylinderScan/Windows+macOS equivalent — read its `design.md` for the pattern this should follow: cross-platform TypeScript verification scripts via Playwright's `_electron` driver + `better-sqlite3`, not a bash-script extension)
- #295 (companion issue: documenting the Linux packaged-app deployment mechanism this CI job would be verifying)
- Proposed as a new tier ("Packaging CI & Linux Deployment Docs") in `docs/superpowers/plans/2026-07-30-graviscan-renderer-roadmap.md`

eberrigan:
[...]

PR #305, opened 2026-08-05
CylinderScan Tier 5a: packaging CI (npm run make + Windows) + hardware doc fixes

### LINKED ISSUE #57: [Testing] Add cross-platform packaged app database testing (Linux, Windows)
## Summary
Extend packaged app database smoke test to run on all platforms (Linux, macOS, Windows).

## Priority
**P3 (Medium-Low)** - Nice to have, but not blocking

## Estimated Effort
6-8 hours (2 days)

## Context
**Current state:**
- ✅ Packaged app smoke test runs on **macOS only**
- ✅ Integration tests run on Linux, macOS, Windows

**Gap:**
- Platform-specific packaging issues possible
- Windows has different path handling (`\` vs `/`)
- Linux might have different ASAR behavior
- Platform-specific Prisma binary issues

**Why macOS only currently:**
- Fastest CI for packaging (M1 runners)
- Most development happens on macOS
- Good enough for smoke test

**Why we need all platforms:**
- Production users on all three platforms
- Platform-specific bugs exist (seen in pilot)
- Windows especially prone to path issues

### macOS (Currently Working ✅)
```bash
APP_PATH="out/Bloom Desktop-darwin-arm64/Bloom Desktop.app/Contents/MacOS/Bloom Desktop"
```
- POSIX paths
- Case-sensitive filesystem (usually)
- App bundle structure

### Linux (Needs Implementation)
```bash
APP_PATH="out/Bloom Desktop-linux-x64/bloom-desktop"
```
- POSIX paths
- Needs xvfb for headless execution
- Different executable structure (no .app bundle)
- May need different DISPLAY handling

### Windows (Needs Implementation)
```powershell
APP_PATH="out\Bloom Desktop-win32-x64\Bloom Desktop.exe"
```
- Backslash path separators
- `.exe` extension
- Different environment variables
- PowerShell vs bash scripting

### Update `scripts/test-package-database.sh`

Make fully cross-platform compatible:

```bash
#!/bin/bash
# Test Packaged App Database Integration (Cross-Platform)

set -e

TIMEOUT=30
LOG_FILE="/tmp/bloom-package-test.log"
PLATFORM=$(uname -s)

# Platform-specific app path
case "$PLATFORM" in
  Darwin)
    APP_PATH="out/Bloom Desktop-darwin-arm64/Bloom Desktop.app/Contents/MacOS/Bloom Desktop"
    ;;
  Linux)
    APP_PATH="out/Bloom Desktop-linux-x64/bloom-desktop"
    ;;
  MINGW*|MSYS*|CYGWIN*)
    APP_PATH="out/Bloom Desktop-win32-x64/Bloom Desktop.exe"
    ;;
  *)
    echo "❌ Unsupported platform: $PLATFORM"
    exit 1
    ;;
esac

echo "Platform: $PLATFORM"
echo "App path: $APP_PATH"

# Check if app exists
if [ ! -f "$APP_PATH" ] && [ ! -d "$APP_PATH" ]; then
  echo "❌ Packaged app not found at: $APP_PATH"
  exit 1
fi

# Platform-specific launch
if [ "$PLATFORM" = "Linux" ]; then
  # Linux: Use xvfb for headless
  xvfb-run --auto-servernum "$APP_PATH" > "$LOG_FILE" 2>&1 &
else
  # macOS, Windows: Direct launch
  "$APP_PATH" > "$LOG_FILE" 2>&1 &
fi

APP_PID=$!

# ... rest of test logic (same for all platforms) ...
```

### Alternative: PowerShell Script for Windows

Create `scripts/test-package-database.ps1` for Windows:

```powershell
# Test Packaged App Database Integration (Windows)

$Timeout = 30
$LogFile = "$env:TEMP\bloom-package-test.log"
$AppPath = "out\Bloom Desktop-win32-x64\Bloom Desktop.exe"

Write-Host "Windows Database Test"
Write-Host "====================="

if (-not (Test-Path $AppPath)) {
    Write-Host "❌ Packaged app not found at: $AppPath"
    exit 1
}

# Launch app
$Process = Start-Process -FilePath $AppPath -RedirectStandardOutput $LogFile -RedirectStandardError $LogFile -PassThru -NoNewWindow

# Wait for initialization
$Elapsed = 0
$Initialized = $false

while ($Elapsed -lt $Timeout) {
    if (Select-String -Path $LogFile -Pattern "\[Main\] Database initialized and handlers registered" -Quiet) {
        $Initialized = $true
        break
    }
    
    if (Select-String -Path $LogFile -Pattern "\[Main\] Failed to initialize database" -Quiet) {
        Write-Host "❌ Database initialization failed"
        Get-Content $LogFile -Tail 20
        Stop-Process $Process
        exit 1
    }
    
    Start-Sleep -Seconds 1
    $Elapsed++
}

Stop-Process $Process -ErrorAction SilentlyContinue

if ($Initialized) {
    Write-Host "✅ SUCCESS"
    exit 0
} else {
    Write-Host "❌ FAILURE: Timeout"
    exit 1
}
```

## Implementation Tasks

- [ ] Update `scripts/test-package-database.sh` for cross-platform:
  - [ ] Add platform detection
  - [ ] Add platform-specific app paths
  - [ ] Add xvfb handling for Linux
  - [ ] Test on Linux manually
  - [ ] Test on macOS (ensure still works)
- [ ] Create `scripts/test-package-database.ps1` for Windows:
  - [ ] PowerShell equivalent logic
  - [ ] Windows-specific paths
  - [ ] Test on Windows manually
- [ ] Update `package.json` scripts:
  ```json
  "test:package:database": "bash scripts/test-package-database.sh",
  "test:package:database:windows": "powershell -File scripts/test-package-database.ps1"
  ```
- [ ] Update CI workflow for matrix:
  ```yaml
  test-package-database:
    name: Test - Packaged App Database (${{ matrix.os }})
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
    
    steps:
      - uses: actions/checkout@v5
      - uses: actions/setup-node@v6
      - uses: astral-sh/setup-uv@v7
      - run: npm ci
      - run: npx prisma generate
      - run: npm run build:python
      - run: npm run package
      
      - name: Test database (Linux)
        if: runner.os == 'Linux'
        run: |
          sudo apt-get install -y xvfb
          npm run test:package:database
      
      - name: Test database (macOS)
        if: runner.os == 'macOS'
        run: npm run test:package:database
      
      - name: Test database (Windows)
        if: runner.os == 'Windows'
        run: npm run test:package:database:windows
  ```
- [ ] Update documentation

## Success Criteria
- ✅ Test script works on macOS, Linux, Windows
- ✅ CI runs successfully on all platforms
- ✅ Platform-specific issues caught (paths, binaries, etc.)
- ✅ Test completes within 30s per platform (~90s total)

## CI Impact
- **Time:** ~30 seconds per platform = ~90 seconds total
- **Platforms:** All three (Linux, macOS, Windows)
- **When:** Every PR OR only main branch (configurable based on CI time budget)
- **Optimization:** Could run only on changed platforms

## Dependencies
- **Requires:** #53 (database infrastructure with smoke test)
- **Optional:** #55, #56 (dev mode and full ops tests, similar patterns)

## Related Issues
- #53 - Scanner-database integration (has macOS-only smoke test)
- #55 - Dev mode database test
- #56 - Full operations test
- #22 - Multi-platform build matrix (related CI infrastructure)

## Platform-Specific Issues to Watch

**Linux:**
- xvfb setup
- Different Prisma binary (libquery_engine-linux-arm64-openssl-3.0.x.so.node)
- Permissions on unpacked files

**Windows:**
- Path separators (`\` vs `/`)
- Prisma binary (query_engine-windows.dll.node)
- PowerShell vs CMD scripting
- Windows Defender might flag binaries

**macOS:**
- Code signing (for distribution, not CI)
- Different architectures (x64 vs arm64)
- App bundle vs flat executable

## Next Steps
After this issue:
- Might reveal platform-specific bugs to fix
- Provides confidence for multi-platform releases
- Builds foundation for #22 (multi-platform builds)

### LINKED ISSUE #180: Hardware validation tests: CylinderScan (Basler + NI-DAQ) — audit existing docs
## Context

CylinderScan has hardware testing docs (`docs/CAMERA_TESTING.md`, `docs/DAQ_TESTING.md`, `docs/SCANNER_TESTING.md`) and integration tests (`tests/integration/test-camera.ts`, `test-daq.ts`, `test-scanner.ts`, `test-scanner-database.ts`). However, these were written during initial development and may be outdated as the codebase has evolved through Phases 2-3.

### Documentation accuracy

- [ ] `docs/CAMERA_TESTING.md` — verify camera IPC commands, settings ranges, and troubleshooting steps match current code
- [ ] `docs/DAQ_TESTING.md` — verify DAQ IPC commands, position tracking, and turntable settings match current code
- [ ] `docs/SCANNER_TESTING.md` — verify scanner workflow (camera+DAQ coordination), frame count calculations, and metadata capture match current code
- [ ] `docs/CONFIGURATION.md` — verify all config fields documented, defaults match `config-store.ts`
- [ ] `.claude/commands/hardware-testing.md` — verify environment variables, mock flags, and troubleshooting steps are current

### Integration test coverage

- [ ] `test-camera.ts` — does it cover current camera settings (gain changed from int 100 to float, brightness added)?
- [ ] `test-scanner.ts` — does it cover `metadata.json` writing (added in recent PR)?
- [ ] `test-scanner-database.ts` — does it cover `experiment_type` field (added in Increment 1)?
- [ ] Are there tests for scanner identity service (added recently)?
- [ ] Are there tests for idle session reset behavior?

### Missing documentation

- [ ] No `docs/TROUBLESHOOTING.md` exists — troubleshooting is scattered across 4 docs. Consider creating a unified troubleshooting guide.
- [ ] No bench test checklist for pre-release hardware validation (manual steps for verifying real Basler camera + NI-DAQ turntable)
- [ ] Packaged app hardware testing procedure — `docs/PACKAGED_APP_TESTING.md` covers DB but not camera/DAQ/scanner in packaged builds

### Metadata verification

- [ ] Verify `metadata.json` fields match current `ScanMetadataJson` interface
- [ ] Verify scan directory format matches `docs/PILOT_COMPATIBILITY.md`
- [ ] Confirm metadata round-trip: capture scan → read metadata.json → all fields present and correct

## Priority

Medium — existing tests pass in CI, but docs may mislead developers. Should be done before the next hardware-connected release.

## References

- Existing docs: `docs/CAMERA_TESTING.md`, `docs/DAQ_TESTING.md`, `docs/SCANNER_TESTING.md`
- Integration tests: `tests/integration/`
- Issue #179 (GraviScan hardware validation — parallel effort)

### LINKED ISSUE #251: CI gap: fully-packaged app (make/DMG + Finder-launch) not exercised end-to-end
## Summary

CI packages and smoke-launches the app, but there is a seam where a regression that breaks **only the fully-packaged, installed app** (specifically the `make`/DMG output or the real Finder-launch path) could pass all checks. This surfaced while building a DMG from `main` for a demo recording.

## What CI covers today

- **`test-package-database`** (`.github/workflows/pr-checks.yml:549`, macOS-only): runs `npm run package`, then `npm run test:package:database` → `scripts/test-package-database-full.sh`. That script launches the packaged binary directly (`out/Bloom Desktop-darwin-arm64/Bloom Desktop.app/Contents/MacOS/Bloom Desktop`) and waits for the `[Main] Database initialized and handlers registered` log line. So a hard main-process early-quit **is** caught here — on macOS.
- **`test-e2e-dev`** (`.github/workflows/pr-checks.yml:435`): Playwright drives the app via `npm run start` (webpack **dev build** through `electron-forge start`), **not** the packaged `.app`.

## The gaps

1. **`make` / DMG maker is never exercised.** CI only runs `npm run package`. The `MakerDMG`/`MakerZIP`/`MakerSquirrel` steps in `forge.config.ts` never run in CI, so a broken installer (bad DMG layout, maker config regression, asar packing that only differs under `make`) ships undetected. `package` and `make` share most of the pipeline but not the maker stage.
2. **No test launches the packaged app the way a user does** (mount DMG → drag to `/Applications` → Finder launch). `test-package-database` invokes the inner Mach-O binary directly. That is close, but skips: DMG mount/copy integrity, Gatekeeper/quarantine behavior of an ad-hoc-signed bundle, and the LaunchServices path.
3. **E2E asserts behavior only on the dev build.** Renderer/IPC flows are green in dev, but the packaged bundle (asar contents, `extraResource` resolution, resource paths under `process.resourcesPath`) is never driven end-to-end past DB init. There is no assertion that a **window actually renders** in the packaged app — the DB smoke test checks a log line, not a visible window.
4. **Packaged-app launch is only verified on macOS.** `test-package-database` is `runs-on: macos-latest` only. Windows (Squirrel) and Linux (deb/rpm) packaged launches are unverified.

> Note: the immediate thing that blocked the demo was `ELECTRON_RUN_AS_NODE=1` set in a local shell (a test-env artifact, not an app bug — the app launches fine from Finder). But chasing it exposed these real coverage gaps.

## Proposed follow-ups

- [ ] Add a **packaged-app smoke test** that runs `npm run make`, mounts the DMG (or installs the platform artifact), launches from the installed location, and asserts the process survives N seconds **and a window renders** (not just a DB log line). Extend `scripts/test-package-database-full.sh` or add a sibling.
- [ ] Exercise `make` (not just `package`) in `test-package-database` (or a new `test-make` job) so the DMG/Squirrel/deb makers run in CI.
- [ ] Extend packaged-app launch verification to **Windows and Linux** runners.
- [ ] Consider a minimal **Playwright smoke run against the packaged `.app`** (not just the dev build) to catch resource-path/asar regressions that only appear in production packaging.

## References

- `.github/workflows/pr-checks.yml:549` — `test-package-database` (package, macOS-only)
- `.github/workflows/pr-checks.yml:435` — `test-e2e-dev` (dev build)
- `scripts/test-package-database-full.sh:131` — direct-binary launch + DB-log wait
- `forge.config.ts:34` — makers (DMG/ZIP/Squirrel/deb/rpm) never run in CI

eberrigan:
## Summary

Tier 5a of the CylinderScan finalization roadmap. OpenSpec change: `add-cylinderscan-packaging-ci`.

- Adds `test-make` (macOS) and `test-make-windows` CI jobs that run `npm run make` end-to-end for the first time — asserting the DMG/Squirrel maker artifacts actually exist and are non-trivially sized, then launching the staged app binary and asserting a real window renders and the app's own database connection/IPC handlers actually work. This is CylinderScan's first Windows packaged-app verification. Closes #251, closes #57.
- Both new jobs upload the verified installer artifacts as downloadable CI build artifacts (7-day retention) so they can be manually test-installed ahead of Tier 5b's hardware QA.
- Fixes stale/incorrect hardware documentation (`CAMERA_TESTING.md`, `DAQ_TESTING.md`, `SCANNER_TESTING.md`, `CONFIGURATION.md`, `hardware-testing.md`) to match already-accepted specs, documents the previously-undocumented scanner identity service and idle-session-reset behavior, and consolidates troubleshooting content into a new `docs/TROUBLESHOOTING.md`. Closes #180.
- Re-wires the orphaned `tests/integration/test-camera.ts` as `test:camera:connect`.

**Two small application-code fixes, both found by actually running the new verification against a real packaged Windows build, not by reading code:**

1. `database.ts`'s Prisma-client symlink fails with `EPERM` on Windows without Developer Mode/admin privilege — the default state for most accounts, including GitHub-hosted runners' non-elevated processes. This is a genuine, pre-existing bug (no CI job has ever launched a packaged app on Windows before this change) that would have made the new Windows job fail on every future PR. Fixed via a new, unit-tested `ensureSymlinkOrCopy()` helper (`src/main/fs-symlink-or-copy.ts`) that falls back to a real recursive copy when the symlink attempt fails.
2. The first version of the DB-verification check (window renders + on-disk schema intact) turned out to be structurally unable to catch that exact bug — the schema is seeded externally *before* launch, so the file-based check would pass identically whether the app's own DB init succeeded, failed, or never ran. Found during this PR's own post-implementation review, then **confirmed empirically** by deliberately reintroducing the bug and re-running the check (it passed when it shouldn't have). Fixed by adding a real IPC round-trip through the running app's renderer (`window.electron.database.phenotypers.list()`), which requires the DB handlers to have actually registered and the connection to actually work — re-confirmed this correctly fails when the bug is reintroduced, and passes when it's fixed. See `design.md` Decision 2d.

A third instance of the same symlink-EPERM bug class was found in `box-backup.ts` (unguarded `symlinkSync`, untested) — deliberately **not** fixed here since it stages potentially large files repeatedly rather than a one-time small copy, and deserves its own design decision. Filed as #303.

**Verified on real CI, not just locally**: both `Test - Make (Windows)` and `Test - Make (macOS)` passed on the actual GitHub-hosted runners on the first push (11m47s / 6m21s).

## Follow-up work (filed, not in this PR)

- #293 — audit GraviScan's own hardware-validation docs (this PR only covers CylinderScan).
- #294, #295 — GraviScan's Linux packaging-CI equivalent + documenting its Linux-only `libusb-filter.so`/`LD_PRELOAD` deployment mechanism. Tracked as a new Tier 6 in the GraviScan renderer roadmap.
- #296 — an actual release pipeline (versioning, signing, publish target) for both platforms.
- #303 — `box-backup.ts`'s same-class symlink bug, untested, needs its own fallback design.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test:unit` — 1264 passed, 7 pre-existing failures confirmed identical on clean `main` via `git stash` (unrelated to this change), 27 new tests all passing
- [x] `npm run format:check` — clean except 6 pre-existing unrelated files (confirmed identical on clean `main` and on recent `main` CI runs — a separate branch `fix-lint-node-formatting-drift` already tracks this)
- [x] `npx openspec validate add-cylinderscan-packaging-ci --strict` — valid
- [x] Local dry-run of both new verification scripts against a real `npm run make` Windows build, including a deliberate regression test (reintroducing the Prisma symlink bug) to prove the new IPC-based DB check actually catches it
- [x] Real CI run on this PR — `Test - Make (Windows)` and `Test - Make (macOS)` both passed on the actual `windows-latest`/`macos-latest` GitHub-hosted runners
- [x] Post-implementation 5-lens subagent `/review-pr` — found and fixed one structural gap in the DB-verification check (see above), verified empirically before and after the fix


eberrigan (commented):

## Review Summary

Five-lens subagent review of the initial implementation (commit 106eae7) found one genuinely important structural gap and several smaller drift/coverage issues; no security or spec-mismatch blockers. The important finding was verified empirically (not just reasoned about) by deliberately reintroducing the bug it describes, confirmed the gap was real, and fixed it in a follow-up commit (b14fa83) — also verified empirically.

## Blocking Issues (found and fixed before this review was posted)

**The DB-verification check was structurally unable to catch the exact bug it exists to guard against.** `applySchemaExternally()` seeds a fully valid schema *before* the app launches; the app's own `initializeDatabaseAsync()` never touches that file (it's lazy — constructs the client but never connects). Because of that, checking the file's schema *after* launch would pass identically whether the app's own database code path succeeded, failed, or never ran — `main.ts` creates the window and continues regardless of DB-init outcome. **Confirmed empirically**: temporarily reverted `database.ts` to the pre-fix buggy symlink-only code, rebuilt, and re-ran `test-packaged-app-launch.ts` — the window still rendered and the file-only check still passed. Fixed in b14fa83 by adding `verifyDatabaseIpcWorks()`, which calls the existing read-only `window.electron.database.phenotypers.list()` IPC method from inside the running renderer via `page.evaluate()` — this requires `registerDatabaseHandlers()` to have actually run and the Prisma connection to actually work. Re-ran the same regression with the fix in place: it now correctly fails with `"No handler registered for 'db:phenotypers:list'"`. Reverted back to the real fix and reconfirmed the full script passes end-to-end. Full writeup: `design.md` Decision 2d.

## Important Issues (found and fixed)

- `resolveDbPath()` used `process.env.HOME || process.env.USERPROFILE`, but the app resolves its own DB path via Electron's `app.getPath('home')` (OS-native, never consults `HOME` on Windows) — a shell environment (e.g. Git Bash, used throughout this implementation) can set `HOME` to a value that diverges from the real profile dir, silently pointing the check at a path the app never writes to. Switched to `os.homedir()`.
- `tests/integration/lib/verify-db-tables.ts`'s `EXPECTED_TABLES` was missing `GraviExperimentWaveMetadata` (added to `prisma/schema.prisma` after the bash-script list this was ported from was last updated) — added it.
- `docs/PACKAGING.md` had a stale code sample and prose describing the pre-fix symlink-only behavior with no copy fallback — updated both.
- `scripts/test-package-database-full.sh:232`'s log-grep for the old `"[Database] Created symlink"` message would have silently stopped matching once the fix changed the log prefix to `[fs-symlink-or-copy]` — updated the grep (cosmetic-only; doesn't change what that job asserts or its pass/fail behavior).

## Also verified against real CI (not just local)

- `Test - Make (Windows)` and `Test - Make (macOS)` both **passed** on the actual GitHub-hosted runners (11m47s / 6m21s) on the first push — confirming the core approach (Playwright against a packaged binary, artifact verification, migration step) works on real infrastructure, not just this dev machine. (A local-machine-only crash surfaced during additional dry-running after the fix commit — traced to something specific to this dev box, not reproduced on the real runner; not a code issue.)
- `Lint - Node.js`'s `format:check` step fails on this PR — but confirmed via `git stash` and via `gh run list --branch main` that the same 6 files fail identically on `main` itself (last 3 main runs all show this failure), and there's already a separate branch (`fix-lint-node-formatting-drift`) tracking it. Pre-existing, unrelated to this PR, not fixed here.

## Suggestions (not blocking)

- This repo now builds and uploads an unsigned, runnable installer from every PR (including forks) as a public 7-day CI artifact — a new capability worth a conscious sign-off given this repo is public, not a silent side effect.
- `ensureSymlinkOrCopy()`'s `mkdirSync` call is no longer wrapped in the swallow-and-continue try/catch the old code had (arguably a correctness improvement, but untested as its own behavior change).
- Consider giving `electron.launch()` an explicit timeout override rather than relying on Playwright's 30s default, for clarity.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), synthesized and one BLOCKING finding fixed + empirically re-verified before posting.*

Issue #318, opened 2026-08-07
Log a warning for unrecognized Python subprocess stdout lines (defense-in-depth)


eberrigan:
## Context

While investigating #316 (camera streaming IPC responses silently corrupted by concurrent stdout writes), code review noted that `PythonProcess.parseLine()` (`src/main/python-process.ts`) already emits a `'raw'` event for any stdout line matching none of the protocol's recognized prefixes (`STATUS:`, `ERROR:`, `DATA:`, `IMAGE:`, and `CameraProcess`'s `FRAME:`/`TRIGGER_CAMERA`/`IMAGE_PATH `/`IMAGE `), but nothing consumes that event by default. This is part of why #316's bug was invisible until it manifested as an opaque 180-second CI timeout — there was no log line anywhere saying "an unrecognized line arrived."

This is not a fix for #316 (that's fully closed at its source by the stdout lock in #316's fix) — it's independent, speculative hardening: a safety net in case a similar unsynchronized-writer defect is introduced elsewhere in the future.

## Proposal

Tracked by OpenSpec change `add-unrecognized-protocol-line-warning`: register a permanent internal `'raw'` listener in `PythonProcess`'s constructor that `console.warn`s a truncated preview of any unrecognized line, except lines matching a small allowlist of already-legitimate, pre-existing informal output (`WARNING:`/`INFO:` prefixes, and specific known benign prefix-less lines like `camera_mock.py`'s "Generating synthetic test patterns instead") — avoiding false-positive noise on every CI/dev run.

Confirmed GraviScan is unaffected: `ScannerSubprocess` doesn't extend `PythonProcess` and has its own separate stdout parser, so this only applies to CylinderScan's `PythonProcess`/`CameraProcess`.

PR #320, opened 2026-08-09
fix(camera): serialize stdout writes to fix streaming IPC race condition

### LINKED ISSUE #316: Camera streaming IPC responses can be silently corrupted by concurrent stdout writes (macOS CI: Command timeout)
## Problem

CI run [31203784171](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/actions/runs/31203784171) (PR #290) failed "Test - Integration (macos-latest)" in `npm run test:camera` → `npm run test:streaming` with `Error: Command timeout` at `src/main/python-process.ts:269` (the 180s `COMMAND_TIMEOUT_MS` in `sendCommand()`).

Confirmed not caused by PR #290 (`git log` shows zero commits touching `src/main/python-process.ts` or `tests/integration/test-streaming.ts` on that branch) — the bug predates it and lives on `main`.

The job log shows the hang is in the very first `await cameraProcess.startStream(...)` call: there is no `[RESPONSE]`, `[PASS] Stream started successfully`, `[TEST 2]`, or `[TEST 3]` anywhere between `[TEST 1] Starting stream...` and the failure, even though the background streaming thread keeps emitting real frames the whole time (driven by persistent event listeners, not the test's main `async` flow).

## Root Cause

`python/ipc_handler.py`'s stdout writers (`send_status`, `send_error`, `send_data`, `send_frame`, plus several bare `print(...)` calls in `detect_cameras()` and in `python/hardware/camera.py`/`camera_mock.py`/`daq.py`/`daq_mock.py`/`scanner.py`) all write to the shared `sys.stdout` with no synchronization between the main IPC command thread and the `streaming_worker()` background thread.

A `FRAME:` line's base64 JPEG payload can be up to ~266 KB — far larger than a pipe's atomic-write capacity — so flushing one takes multiple OS `write()` syscalls. CPython releases the GIL during each blocking write, so a concurrent write from the main thread (e.g. `start_stream`'s own success response, issued right as the just-spawned thread emits its first frame) can land in the gap between two of those syscalls, corrupting the line-delimited protocol. The short response gets silently folded into the still-open `FRAME:` line and is never recognized as its own line on the Node side — it doesn't throw or reject, it just vanishes, and `sendCommand()`'s promise sits until the 180s timeout fires.

This is distinct from the race fixed by #47 (response/request correlation for *complete, valid* responses arriving out of order) — this corrupts the bytes before a line is ever completed, so correlation never gets a chance to run.

## Fix

Tracked by OpenSpec change `fix-camera-streaming-stdout-race`: a single shared, lock-protected `write_line()` primitive (`python/protocol_io.py`) used by every stdout writer reachable while a camera stream may be active.

Confirmed **not** reachable from GraviScan's scan worker (`python/graviscan/scan_worker.py` has no background thread at all) — this is purely a CylinderScan camera-streaming bug.

eberrigan:
## Summary

Fixes a stdout byte-interleaving race between the camera-streaming background thread and the main IPC thread that caused `sendCommand()` promises to hang until the 180s timeout — observed as a CI "Command timeout" failure ([run 31203784171](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/actions/runs/31203784171) on PR #290).

## Changes

- Add `python/protocol_io.py`: a single shared `threading.Lock()` + `write_line()` primitive that serializes every stdout write reachable while a camera stream may be active.
- Reimplement `python/ipc_handler.py`'s `send_status()`/`send_error()`/`send_data()`/`send_frame()` and `detect_cameras()`'s diagnostic prints to go through `write_line()` instead of bare `print()`.
- Route the same call sites in `python/hardware/camera.py`, `camera_mock.py`, `daq.py`, `daq_mock.py`, `scanner.py` through `write_line()` (excluding each file's standalone-script-only paths and `scanner.py`'s stderr writes — different OS pipe, can't interleave with stdout).
- Add `protocol_io`/`python.protocol_io` to `python/main.spec`'s `hiddenimports` so the fix works in the packaged PyInstaller executable, not just dev-mode.
- Add `pytest-timeout` dev dependency so a future lock-ordering regression fails fast instead of hanging the test suite.

## OpenSpec Change

- Change ID: `fix-camera-streaming-stdout-race`
- Affected capabilities: `ipc-reliability`
- Delta types: ADDED
- All `tasks.md` items complete: yes, except 4.3 (this PR's CI run) and 4.4 (pre-merge sweep, running now) and 5.2 (this description)

## Root Cause

`streaming_worker()`'s background thread emits `FRAME:` lines (base64 JPEG, up to ~266KB) that need multiple OS `write()` syscalls to flush; CPython releases the GIL during each one. The main thread's own command responses (e.g. `start_stream`'s success ack) could land in that gap and get silently folded into the still-open `FRAME:` line — never recognized as a complete protocol line, so the response just vanishes and `sendCommand()` waits out the full `COMMAND_TIMEOUT_MS`.

Re-reading the original CI job log line-by-line showed the hang was actually in `cameraProcess.startStream()` (not `stopStream()`, as originally assumed) — there's no `[RESPONSE]`/`[PASS] Stream started`/`[TEST 2]` marker anywhere after `[TEST 1] Starting stream...` in the failing run.

Confirmed **not** reachable from GraviScan: its scan worker (`python/graviscan/scan_worker.py`) has no background thread at all — this is purely a CylinderScan camera-streaming bug.

### Python Tests

- [x] Unit tests pass (`uv run pytest python/`) — 278 passed, 6 pre-existing/unrelated failures
- [!] 6 pre-existing Windows-only file-handle `PermissionError`s in `test_camera_streaming.py`/`test_scan_worker.py`/`test_tiff_metadata.py` — confirmed via `git diff --stat` that these files are byte-for-byte unmodified by this PR; won't manifest on CI's Linux unit-test job
- [x] New tests added: `python/tests/test_protocol_io.py` (deterministic interleaving regression tests using an instrumented fake stdout, not timing-based), plus `test_stop_stream_returns_promptly_while_streaming` in `test_camera_streaming.py`
- [x] Coverage: 89.80% (≥80% required); `protocol_io.py` itself at 100%

### Integration Tests

- [x] Camera tests pass locally against the real packaged binary (`npm run test:camera`) — `ALL TESTS PASSED`, with the previously-missing `[RESPONSE]`/`[PASS] Stream started successfully` now appearing immediately after `[TEST 1] Starting stream...`

## Build & Packaging

- [x] Python executable builds successfully (`npm run build:python`) — confirms `protocol_io` resolves via the new `hiddenimports` entries

## Linting & Formatting

- [x] Python formatting applied (`uv run black python/`)
- [x] Python linting passes (`uv run ruff check python/`)
- [x] Python type checking passes (`uv run mypy python/`) — only the pre-existing, unrelated `scan_worker.py:592` error remains

- [ ] CI tests pass on Linux, macOS, and Windows — pending this PR's CI run (this is the primary thing this PR needs CI to confirm, since the original bug was macOS-specific timing)

## Related Issues

Fixes #316

## Reviewer Notes

This PR was reviewed through two rounds of a 5-agent adversarial OpenSpec review before implementation (spec quality, code/architecture feasibility, GitHub issue alignment, TDD strategy, scientific rigor) — see `openspec/changes/fix-camera-streaming-stdout-race/` for the full proposal/design rationale, including why the lock scope had to expand beyond the original 4 helper functions, and why `_stdout_lock` must stay separate from the existing `_streaming_lock`.

A companion, independent change (`add-unrecognized-protocol-line-warning`, tracked by #318) adds Node-side defensive logging for unrecognized protocol lines — deliberately not bundled into this PR since it's speculative hardening, not part of this bug's root-cause fix.


eberrigan (commented):

## Review Summary

Five-agent adversarial review (code quality, testing/TDD, scientific rigor, security/cross-platform, behavioral correctness) found **no blocking issues**. The root-cause fix is sound: `_stdout_lock` genuinely covers every reachable stdout writer, the separation from `_streaming_lock` is correctly implemented (not just documented), and no reentrancy/deadlock path exists. One real gap was found and **fixed during this review**: `test_hardware_diagnostic_does_not_interleave_with_frame` passed even with the lock deliberately removed, because `MockCamera`'s constructor does real filesystem work before reaching `write_line()`, unlike the other two interleaving tests where thread B's entire body *is* the write. Added a `write_entered` synchronization signal and empirically re-verified: all three interleaving tests now fail deterministically across repeated runs with the lock stripped, and pass deterministically with it restored.

## Fixed During Review

- **Real test gap**: `test_hardware_diagnostic_does_not_interleave_with_frame` didn't reliably exercise the race it claimed to test (see above) — fixed with a proper synchronization primitive, verified by manually stripping/restoring the lock 3x each.
- **Doc-precision**: `design.md`/`tasks.md` said "six" import-time prints in `ipc_handler.py`; actual count is seven. Fixed, plus documented why `daq.py`'s own import-time print was routed through `write_line()` while `ipc_handler.py`'s weren't (intentional asymmetry, not an oversight).
- **Overclaiming docstring**: `test_stop_stream_returns_promptly_while_streaming`'s comment claimed its event hook "genuinely guarantees" landing mid-write; it's biased toward that, not guaranteed (unlike the `gap_open`/`resume` handshake). Softened to match what the mechanism actually provides.
- Added a **Known Limitations** section to `design.md`: SIGTERM/kill mid-write is unaddressed (acceptable — no deadlock, just abandoned I/O, same as any in-flight write today), and the actual race-condition unit tests (`test_protocol_io.py`) run Ubuntu-only in CI (`Test - Python` isn't platform-matrixed) — `Test - Integration` does cover all 3 platforms and would have caught this exact regression, just less directly. Matrixing `Test - Python` is a bigger, repo-wide CI cost decision noted as a follow-up candidate, not done here.

## Notes

- Pre-existing, unrelated `Lint - Node.js` Prettier-drift failure (files this PR never touches) — `main` has no branch protection, so it doesn't block merge.
- All three `Test - Integration` platforms pass, including macOS — the exact job that originally failed with `Command timeout` in CI run 31203784171.

---

PR #321, opened 2026-08-09
feat(ipc): warn on unrecognized Python subprocess stdout lines

### LINKED ISSUE #318: Log a warning for unrecognized Python subprocess stdout lines (defense-in-depth)

eberrigan:
## Summary

Adds a default `console.warn` for stdout lines from the Python subprocess that don't match any recognized protocol prefix — a general-purpose safety net discovered while investigating #316, since that bug's dropped response was invisible until it manifested as an opaque 180s timeout.

## Changes

- `PythonProcess`'s constructor registers a permanent internal `'raw'` listener (`registerUnrecognizedLineWarning()`) that `console.warn`s a truncated (~200 char) preview of any line not matching a recognized prefix.
- Allowlists two categories of already-legitimate, pre-existing informal Python output so they don't create false-positive warning noise: `WARNING:`/`INFO:` prefixes (exact, case-sensitive), and an exact-match list currently containing `camera_mock.py`'s prefix-less `"Generating synthetic test patterns instead"` fallback message (a prefix allowlist alone can't match a prefix-less line — this exact gap was caught and initially missed in an earlier draft of the underlying OpenSpec proposal, see `openspec/changes/add-unrecognized-protocol-line-warning/proposal.md`).
- The existing `'raw'` event is unchanged — still fires unconditionally for any explicit listener; this only adds a default warning on top.
- Confirmed GraviScan is unaffected: `ScannerSubprocess` doesn't extend `PythonProcess` and has its own separate stdout parser.

## OpenSpec Change

- Change ID: `add-unrecognized-protocol-line-warning`
- Affected capabilities: `ipc-reliability`
- Delta types: ADDED
- All `tasks.md` items complete: yes, except 3.2 (pre-merge, running now) and 3.3 (this description)

## Relationship to #316 / PR #320

Independent of the stdout-race-condition fix in PR #320 (`fix-camera-streaming-stdout-race`) — this is speculative hardening for a *general* failure class, not a fix for that specific bug, which is fully closed at its source by PR #320's stdout lock. Deliberately kept as its own change/PR (rather than bundled) since it touches a disjoint file (`src/main/python-process.ts` only) with no dependency in either direction, and reviewers flagged that bundling would muddy #320's scope and archival record. Can merge independently, in either order.

### TypeScript Tests

- [x] Unit tests pass (`npm run test:unit`)
- [x] Unit tests added: 7 new tests across `tests/unit/python-process.test.ts` (unrecognized line warns; `WARNING:`/`INFO:`/benign-prefix-less line don't warn; case-sensitivity and false-prefix-trap both warn; empty line doesn't warn) and `tests/unit/camera-process.test.ts` (a `FRAME:` line handled by `CameraProcess`'s own `parseLine()` override never reaches the base warning logic at all)
- [!] 5 pre-existing failures in `config-store.test.ts`, `graviscan/database-handlers.test.ts`, `graviscan/scan-coordinator.test.ts`, `image-uploader.test.ts`, `scans-export.test.ts` — confirmed via `git diff --stat` that none of these files are touched by this PR; pre-existing suite flakiness unrelated to `python-process.ts`/`camera-process.ts`

## Linting & Formatting

- [x] Prettier formatting applied
- [x] TypeScript compiles without errors (`npx tsc --noEmit`)

## Related Issues

Fixes #318
Related to #316 / #320


eberrigan (commented):

## Review Summary

Five-agent adversarial review found **no blocking issues** on this small, well-scoped hardening change. Verified independently (not just from the diff): the constructor registration is correct and singleton-safe (no `EventEmitter` max-listener risk across `restart()` cycles), the allowlist check never runs on the high-frequency `FRAME:` streaming path, `CameraProcess`'s own recognized prefixes correctly never reach the base class's warning logic, and TDD red→green was independently reproduced (temporarily disabling the registration reproduced the exact 3-test failure the PR's own red-step evidence claims).

## Fixed During Review

- **Real test gap**: only `FRAME:` was tested against `CameraProcess`'s 4 recognized prefixes; `TRIGGER_CAMERA`, `IMAGE `, and `IMAGE_PATH ` were untested. Added tests for all three — verified each one genuinely catches a regression (temporarily disabled `TRIGGER_CAMERA` handling, confirmed only its test failed).
- **Fragile test pattern**: `camera-process.test.ts`'s new test grabbed the spawned mock via a hardcoded `mock.results[0]` index, which only worked because no earlier test in the file called `.start()`. Fixed with `mockClear()` + `.at(-1)`.
- **Incomplete assertions**: 2 of 7 tests (case-sensitivity, false-prefix-trap) only checked `console.warn`, not that the existing `'raw'` event still fires — added those assertions.
- **UTF-16 edge case**: `line.slice(0, 200)` could split a surrogate pair (e.g. an emoji) mid-character. Switched to code-point-safe truncation via `Array.from().slice().join()`.
- **Doc gap**: `proposal.md`'s Impact section didn't mention `camera-process.test.ts`'s mock-fidelity change (fresh object per `spawn()` call, needed for the new tests) — documented.

## Significant Finding (Filed Separately, Not Fixed Here)

Review surfaced a **pre-existing, unrelated race condition**: `PythonProcess.restart()`'s `stdoutChunks` buffer isn't generation-scoped, unlike the `'exit'` handler — bytes from a killed process and its replacement could concatenate into one corrupted line during a restart cycle. Structurally the same corruption class as #316, but for the base `restart()` path rather than camera streaming specifically. Filed as **#322**; out of scope for this PR.

## Notes

- Confirmed `console.warn` never reaches the renderer UI (main-process-only logging).
- Allowlist-staleness risk (a future benign prefix not yet allowlisted) is inherent to any allowlist and self-correcting — acceptable, already has a code comment reminding maintainers.
- Same pre-existing, unrelated `Lint - Node.js` Prettier-drift failure as PR #320 — doesn't block merge.

---

Issue #343, opened 2026-08-14
Add Machine Configuration UI for graviscan_system_name, slack_webhook_url, libusb_endpoint_recovery


eberrigan:
## Problem

Three fields in the `~/.bloom/.env` machine-config schema (`MachineConfig` interface in `src/main/config-store.ts`) have no UI in Machine Configuration at all today — they can only be set by manually editing the `.env` file:

- `graviscan_system_name`
- `slack_webhook_url` (a.k.a. `BLOOM_GRAVISCAN_SLACK_WEBHOOK_URL`)
- `libusb_endpoint_recovery`

All three are hydrated into `process.env` once at main-process startup (per their doc comments in `config-store.ts`), same as `scanner_mode` — so even if UI were added for them, changing them would also need the "restart required" notice pattern being introduced for `scanner_mode` in #334.

## Context

Surfaced while scoping #334 (Machine Config's silent-restart-required bug). Investigation confirmed `scanner_mode` is currently the *only* Machine Config field with a real restart-required bug reachable through the UI — these three fields have the same underlying issue but no UI path to trigger it, so they're out of scope for #334's fix.

## Suggested fix

- Add form fields for these three to `MachineConfiguration.tsx` (likely a new "Advanced" or "GraviScan System" section, given two are GraviScan/USB-specific).
- Reuse the restart-required notice mechanism from #334 once these fields are save-able through the UI.
- Consider consolidating with #325 (canonical two-tier `.env` schema doc) and #204 (some manual-only fields, e.g. `GRAVISCAN_MOCK`, aren't loaded into `process.env` at startup at all) since all three issues touch the same underlying config schema gaps.

## Related

- #325 — file a proper two-tier doc for the full `.env` schema
- #204 — `GRAVISCAN_MOCK` and other `.env` values not loaded into `process.env` at main-process startup
- #334 — Machine Configuration changes silently require a full app restart (the notice pattern this issue would reuse once these fields have UI)

Issue #337, opened 2026-08-13
Sidebar nav ordering doesn't match the Daily-Workflow-first UX established on the Home page


eberrigan:
## Problem

The sidebar (`Layout.tsx`'s `alwaysLinks`/`captureLinks`) currently orders CylinderScan-mode links as:

Home, Browse Scans, Export Scans, Scientists, Phenotypers, Experiments, Capture Scan, Camera Settings, Accessions

This doesn't match the "Daily Workflow" (Camera Settings → Capture Scan → Browse Scans) / "Setup" (Scientists, Phenotypers, Accessions, Experiments) structure that `CylinderScanWorkflowGuide.tsx` (PR #329, Tier 4) just established as the sensible task order on the Home page itself — the persistent sidebar and the Home page quickstart guide now disagree with each other on ordering, which is confusing.

Flagged during manual walkthrough of PR #329 as "not intuitive" — pre-existing sidebar structure, not something #329 changed (that PR deliberately left `Layout.tsx` untouched, see its "Deferred Scope").

## Suggested reordering (starting point, not final)

Daily workflow first: Home, Camera Settings, Capture Scan, Browse Scans, Export Scans — then Setup: Scientists, Phenotypers, Accessions, Experiments.

Whoever picks this up should confirm the final order (this issue is just flagging the mismatch, not mandating the exact fix), and check whether GraviScan's link ordering has an analogous issue worth addressing in the same pass.

Issue #335, opened 2026-08-13
Verify accuracy of Basler camera IP-finding instructions on Camera Settings page


eberrigan:
## Problem

`src/components/CameraSettingsForm.tsx` (~line 204-228) shows a help panel with instructions for finding a Basler camera's IP address:

```
Method 1: Basler Pylon Viewer (Recommended)
  - Open Basler Pylon Viewer software
  - Right-click camera → Properties
  - View IP Address in properties panel

Method 2: Check Camera Label
  - Physical label on camera may show IP address

Method 3: Router Admin Page
  - Check connected devices in router settings (usually 192.168.1.1)
```

Nobody currently working on this has physical access to the Basler hardware to verify these steps are accurate or that Pylon Viewer is genuinely the recommended method. Flagged during a manual walkthrough of PR #329 — not a claim that this is wrong, just unverified copy that should be checked by someone with real camera/network access before treating it as reliable.

## Ask

- Verify each method against the actual Basler acA2000-50gm + Pylon Viewer workflow this app targets.
- Correct or remove any step that's wrong.
- Consider whether "Method 3: Router Admin Page" is even a reasonable ask for a non-technical lab user, or should be dropped/rephrased.

eberrigan:
The help-text panel this issue is about ("Method 1: Basler Pylon Viewer / Method 2: Check Camera Label / Method 3: Router Admin Page", currently in `CameraSettingsForm.tsx`) has relocated to Machine Configuration's Hardware section as part of #338 (removing the camera-selection/manual-entry UI from the Camera Settings page). The content itself is unchanged — this issue's ask (verify accuracy against real Basler/Pylon Viewer hardware) still applies, just at its new location.

PR #344, opened 2026-08-17
fix(cylinderscan): config/UX quick fixes from Tier 4 walkthrough

### LINKED ISSUE #333: Update default bloom_api_url from api.bloom.salk.edu/proxy to bloom.salk.edu/api
## Problem

`getDefaultConfig()` in `src/main/config-store.ts:155` hardcodes:

```ts
bloom_api_url: 'https://api.bloom.salk.edu/proxy',
```

This is the default seeded into Machine Configuration for a fresh install (no existing `.env`). The current correct default is **`https://bloom.salk.edu/api`**.

## Fix

Update the default in `getDefaultConfig()`. Also grep for any other hardcoded references to the old URL (docs, `.env.example`, test fixtures) so they don't drift out of sync again.

Found during manual walkthrough of PR #329 (Tier 4 style/UX parity) — unrelated to that PR's actual changes, just noticed while testing Machine Configuration.

### LINKED ISSUE #334: Machine Configuration changes (e.g. Scanner Mode) silently require a full app restart to take effect
## Problem

`useAppMode()` (`src/renderer/hooks/useAppMode.ts`) fetches `scanner_mode` from the main process exactly once, on mount, with an empty effect dependency array — it never re-fetches. `MachineConfiguration.tsx`'s save handler (`handleSubmit`, around line 122) only shows a 3-second "saved" success toast; it doesn't trigger any reload, remount, or re-fetch of app-level state.

Net effect: changing Scanner Mode (or likely other settings that gate top-level routing/rendering) in Machine Configuration appears to succeed, but the app keeps behaving as if nothing changed until the user manually restarts it. There's currently zero indication to the user that a restart is needed.

Confirmed via manual walkthrough of PR #329 (Tier 4 style/UX parity) — not caused by that PR, pre-existing behavior.

## Minimal fix

After a successful save where `scanner_mode` (or other app-shell-affecting fields) changed, show an explicit "Restart the app for this change to take effect" notice instead of (or alongside) the generic "saved" toast.

## Bigger fix (separate follow-up, not required here)

Consider making `useAppMode()` re-fetch on a config-changed event (IPC push from main, or a shared store) so mode switches apply live without a restart. That's a larger architectural change and probably shouldn't block the minimal fix above.

### LINKED ISSUE #336: Export page's success message doesn't label what the count is, and silently drops the more-useful exportedScans count
## Problem

`Export.tsx` (~line 328) renders:

```
{resultBanner.exportedFiles} exported, {resultBanner.skippedFiles} skipped (already exist)
```

e.g. "73 exported, 0 skipped (already exist)" — with no indication that "73" is a count of individual **files** (images + metadata.json across all exported scans), not scans. A user has no way to relate that number back to "how many scans did I actually export."

The backend handler (`scansExport()` in `src/main/database-handlers.ts`, ~line 1268-1346) already computes `exportedScans` (a count of whole scans successfully exported) alongside `exportedFiles`/`skippedFiles`/`failedScans` — but `Export.tsx` only destructures `exportedFiles`/`skippedFiles` (~line 250-258) and never surfaces `exportedScans` at all.

Found during a manual walkthrough of PR #329 — unrelated to that PR, pre-existing on the Tier 3 export page.

## Fix

Surface `exportedScans` in the success message, e.g.:

```
12 scans exported (73 files), 0 files skipped (already exist)
```

No backend change needed — the data already exists in the response, just isn't read.

### LINKED ISSUE #338: Remove Camera Settings page's camera-selection/manual-IP-entry UI — Machine Configuration should be the single source of truth for camera_ip_address
## Problem

`camera_ip_address` is currently settable in two different places:

1. **Machine Configuration** (`MachineConfiguration.tsx`, ~line 579) — a direct text input, saved to `~/.bloom/.env` via `config.set()`.
2. **Camera Settings page** (`CameraSettingsForm.tsx`, via `showCameraSelection={true}`) — a dropdown of auto-detected cameras plus a "Manual Entry..." option (`showManualEntry` state, ~line 46/122) that reveals a second, separate text input for the same field.

Two places to set the same value with no clear indication of which one wins or how they reconcile is confusing. Flagged during manual walkthrough of PR #329.

## Requested change

Per user request: remove the camera-selection/manual-entry UI from the Camera Settings page entirely. `camera_ip_address` should be configured only via Machine Configuration.

## Open question for whoever picks this up

Camera Settings' auto-detection dropdown (listing cameras actually found on the network) is arguably more useful than Machine Config's plain text field, since it doesn't require knowing the IP in advance — worth deciding whether that detection capability should be preserved (e.g. moved into Machine Configuration itself, replacing its plain text input) rather than dropped entirely, or whether Machine Config's simple text field is sufficient and detection just goes away. This issue doesn't prescribe which; just that Camera Settings should stop being a second place to set this value.

### LINKED ISSUE #339: Move Check Hardware / Restart Python to Machine Configuration; Home should just show a status indicator
## Problem

`PythonStatus.tsx` renders both a hardware/status display *and* two action buttons — "Check Hardware" and "Restart Python" — directly on the Home page, visible to every user in both scan modes. These are troubleshooting/admin actions, not something a regular lab technician should need (or be tempted) to trigger during normal use.

Flagged during manual walkthrough of PR #329 (Tier 4 style/UX parity) — pre-existing placement, not something #329 changed.

## Requested change

- Move the "Check Hardware" and "Restart Python" actions into Machine Configuration (which already has an admin-only precedent — see closed issue #49, "Add Machine Configuration UI with admin-only access").
- Home page keeps a simple status indicator (connected/disconnected/error) derived from the same underlying state, with no interactive troubleshooting controls — just a glance-and-go status for the person actually scanning.

## Scope notes for whoever picks this up

- Decide whether "Restart Python" needs a confirmation step given it interrupts the running Python subprocess — check whether an in-progress scan should block/warn before allowing a restart.
- Preserve the existing mode-gating (`PythonStatus.tsx`'s `mode !== 'cylinderscan'` early return, or the mode-aware check-hardware/DAQ suppression from #198) when relocating.

eberrigan:
## Summary

Bundles five small, independent, pre-existing UX/config gaps found during a manual walkthrough of PR #329 (Tier 4 style/UX parity) — unrelated to what that PR changed, just noticed while exercising the app. Full design rationale in `openspec/changes/fix-cylinderscan-config-ux-quickfixes/`.

- **Fix stale `bloom_api_url` default** (`https://api.bloom.salk.edu/proxy` → `https://bloom.salk.edu/api`) in both places it was hardcoded, plus docs/scripts/the live spec.
  Closes #333
- **Restart-required notice** when Scanner Mode changes in Machine Configuration — a persistent, dismissible banner instead of the generic auto-dismissing toast. Confirmed `scanner_mode` is the *only* Machine Config field with this bug (every other field is re-read fresh on next use).
  Closes #334
- **Export page's success/partial messages** now surface the scan count alongside the file count (e.g. "12 scans exported (73 files), 0 files skipped"), not just an ambiguous file count.
  Closes #336
- **Camera auto-detection relocated** from the Camera Settings page into Machine Configuration's Hardware section, replacing its plain text input — Machine Configuration is now the sole place to set `camera_ip_address`.
  Closes #338
- **Check Hardware / Restart Python relocated** from Home's status panel into Machine Configuration's Hardware section, with a confirm dialog before restart (no guard against restarting mid-scan existed before). Home now shows only a simple Connected/Checking/Error indicator.
  Closes #339

**Deferred** (not in this PR): #337 (sidebar nav ordering) — PRs #289/#290 are both still open and actively rewriting `Layout.tsx`, same collision Tier 4 deliberately avoided. Follow-up issue #343 filed for Machine Config UI on 3 manual-`.env`-only fields sharing #334's root cause. #335 (Basler camera IP-finding instructions) — content relocates with #338 but isn't verified, since that requires physical hardware access; left a comment noting the move.

**Bug found and fixed via manual E2E smoke-testing** (not part of the original five issues): `main.ts`'s `config:get` IPC handler silently omitted `scanner_mode` from its response. Every unit test covering `MachineConfiguration.tsx` mocks the IPC boundary directly and always included the field, so this was invisible until a real IPC round-trip — in the actual app, the Scanner Mode radios always rendered unchecked and the entire CylinderScan-only Hardware section (including this PR's own new UI) never rendered. Fixed the handler and added a permanent E2E regression test, since this bug class is structurally invisible to IPC-mocking unit tests.

## Test plan

- [x] Proposal reviewed via `openspec-review`'s 5-subagent team (2 rounds to convergence: round 1 found 2 blocking + ~13 important issues, all fixed; round 2 verified all fixes hold, 2 minor items found and fixed)
- [x] TDD throughout: every task in `tasks.md` (26/26 complete) wrote a failing test before implementation
- [x] Full targeted unit/component suite green (233 tests across 9 touched files)
- [x] Full repo unit suite: 1466 passed; remaining failures are pre-existing, Windows-only path-separator/process-management artifacts in files this PR never touches (confirmed via `git log`) — CI runs unit tests on Linux
- [x] `npx tsc --noEmit` and `npm run lint` clean
- [x] Manual smoke test against the real running app (Playwright driving a real Electron instance, isolated `.env`/DB fixture): restart notice, camera detection + manual-entry fallback, Hardware Diagnostics with confirm-gated restart, Camera Settings UI removed, Home's simplified status indicator, Export page — all verified working, with screenshots
- [x] New permanent E2E regression test (`tests/e2e/machine-config-scanner-mode-persistence.e2e.ts`) added for the discovered `scanner_mode` bug, verified red→green by reverting the fix and re-running

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,7 +50,7 @@ NPM_TOKEN=your-gitlab-npm-token-here
 BLOOM_TEST_USERNAME=your-test-username@salk.edu
 BLOOM_TEST_PASSWORD=your-test-password
 BLOOM_ANON_KEY=your-anon-key-here
-BLOOM_API_URL=https://api.bloom.salk.edu/proxy
+BLOOM_API_URL=https://bloom.salk.edu/api
 
 # =============================================================================
 # GRAVISCAN V600 WEDGE FOLLOW-UPS (optional, #228 + #236)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- CylinderScan config/UX quick fixes from Tier 4 walkthrough ([PR #344](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/pull/344), closes [#333](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/333), [#334](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/334), [#336](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/336), [#338](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/338), [#339](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/339))
+  - Fixed stale default `bloom_api_url` (`api.bloom.salk.edu/proxy` → `bloom.salk.edu/api`) in both `config-store.ts` and a second, independent hardcoded copy in `MachineConfiguration.tsx`
+  - Machine Configuration now shows a persistent, dismissible "restart required" notice when Scanner Mode changes, instead of the generic auto-dismissing save toast — confirmed `scanner_mode` is the only Machine Config field with this bug
+  - Export page's success/partial banners now surface the scan count alongside the file count (e.g. "12 scans exported (73 files)"), not just an ambiguous file count
+  - Camera auto-detection relocated from the Camera Settings page into Machine Configuration's Hardware section, replacing its plain text input — Machine Configuration is now the sole place to set `camera_ip_address`
+  - "Check Hardware"/"Restart Python" relocated from Home's status panel into Machine Configuration's Hardware section, with a confirm dialog before restart; Home now shows only a simple Connected/Checking/Error indicator
+  - Found and fixed via manual E2E smoke-testing: `config:get`'s real IPC handler (`main.ts`) silently omitted `scanner_mode` from its response, making the entire CylinderScan-only Hardware section fail to render in the real app despite passing unit tests that mock the IPC boundary directly; added a permanent E2E regression test since this bug class is invisible to mocked unit tests
+  - See `openspec/changes/fix-cylinderscan-config-ux-quickfixes/` for full design rationale; [#337](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/337) (sidebar nav ordering) deferred pending PRs #289/#290; follow-up [#343](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/issues/343) filed for Machine Config UI on 3 manual-`.env`-only fields
 - CylinderScan correctness & security hardening ([PR #280](https://github.com/Salk-Harnessing-Plants-Initiative/bloom-desktop/pull/280), roadmap Tier 1)
   - Replaced `webSecurity: false` with a custom protocol handler + path-traversal validation for local scan-image access (#93)
   - Added cleanup functions to all 8 preload listeners that were missing them (#96)
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -4,6 +4,8 @@ This document describes all configurable constants and default values in the Blo
 
 **For Admin UI Implementation**: When building the admin settings UI, reference this document for all configuration options that should be made user-configurable.
 
+**Accessing Machine Configuration**: The Machine Configuration page has no visible sidebar link — it's reached via the `Ctrl+Shift+,` (`Cmd+Shift+,` on macOS) keyboard shortcut. Since hardware troubleshooting (Check Hardware / Restart Python) now lives there rather than on the Home page, admins should relay this shortcut to lab technicians who may need to report a hardware issue during a scanning session.
+
 ## Scanner Settings
 
 ### Rotation & Timing
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -4,7 +4,7 @@ This document describes all configurable constants and default values in the Blo
 
 **For Admin UI Implementation**: When building the admin settings UI, reference this document for all configuration options that should be made user-configurable.
 
-**Accessing Machine Configuration**: The Machine Configuration page has no visible sidebar link — it's reached via the `Ctrl+Shift+,` (`Cmd+Shift+,` on macOS) keyboard shortcut. Since hardware troubleshooting (Check Hardware / Restart Python) now lives there rather than on the Home page, admins should relay this shortcut to lab technicians who may need to report a hardware issue during a scanning session.
+**Accessing Machine Configuration**: The Machine Configuration page has no visible sidebar link — it's reached via the `Ctrl+Shift+,` (`Cmd+Shift+,` on macOS) keyboard shortcut, intentionally not shared with lab technicians. Since hardware troubleshooting (Check Hardware / Restart Python) now lives there rather than on the Home page, make sure technicians know to contact you (the admin) when Home shows a hardware error — they have no in-app path to these controls themselves, by design.
 
 ## Scanner Settings
 
--- a/docs/MANUAL_UPLOAD_TESTING.md
+++ b/docs/MANUAL_UPLOAD_TESTING.md
@@ -15,7 +15,7 @@ CAMERA_IP_ADDRESS=10.0.0.50
 SCANS_DIR=/path/to/your/scans
 
 # Bloom API Settings (required for upload)
-BLOOM_API_URL=https://api.bloom.salk.edu/proxy
+BLOOM_API_URL=https://bloom.salk.edu/api
 BLOOM_SCANNER_USERNAME=your_scanner@salk.edu
 BLOOM_SCANNER_PASSWORD=your_password
 BLOOM_ANON_KEY=your_anon_key
@@ -152,7 +152,7 @@ You can verify via the Supabase dashboard or using the Supabase CLI.
 
 1. Verify your username/password are correct
 2. Check if your account has access to the Bloom API
-3. Ensure `BLOOM_API_URL` is correct (default: `https://api.bloom.salk.edu/proxy`)
+3. Ensure `BLOOM_API_URL` is correct (default: `https://bloom.salk.edu/api`)
 
 ### "Scan not found"
 
--- a/openspec/changes/fix-cylinderscan-config-ux-quickfixes/design.md
+++ b/openspec/changes/fix-cylinderscan-config-ux-quickfixes/design.md
@@ -1,0 +1,200 @@
+## Context
+
+Five independent, pre-existing UX/config gaps in CylinderScan mode's
+configuration surfaces, all filed as separate GitHub issues (#333, #334,
+#336, #338, #339) after a manual walkthrough of PR #329. Two of them (#338,
+#339) converge on the same section of the same file
+(`MachineConfiguration.tsx`'s Hardware section), so this design covers that
+convergence explicitly rather than treating them as fully independent edits.
+
+## Goals / Non-Goals
+
+- Goals:
+  - Fix each of the five issues completely within its own stated scope.
+  - Make Machine Configuration the single source of truth for
+    `camera_ip_address`, without losing the auto-detection capability.
+  - Give the admin (who now owns hardware troubleshooting) a safety net
+    before restarting Python mid-scan, without building new cross-cutting
+    scan-state-tracking infrastructure.
+- Non-Goals:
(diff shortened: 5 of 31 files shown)


eberrigan (commented):
> **Verdict: REQUEST_CHANGES → fixed in b057b60** (posted as comment — cannot request changes on own PR)

## Review Summary

5-subagent review team (Code Quality, Testing/TDD, Scientific Rigor, Security, Behavioral Correctness) found 4 BLOCKING and ~10 IMPORTANT issues across the diff. All 4 blocking and the majority of important findings are fixed in b057b60. Security review came back clean with no findings.

1. **Basler camera IP-finding help text was deleted, not relocated as design.md promised.** The proposal explicitly said this text moves verbatim into `MachineConfiguration.tsx`; the implementation deleted it from `CameraSettingsForm.tsx` and never added it anywhere. Restored verbatim in the Hardware section.
2. **Real Python crashes were silently misclassified.** `main.ts` pushes `"Process exited: <code>"` via `python:status` on a crash; `PythonStatus.tsx`'s simplified Connected/Error/Checking logic didn't recognize this string, so a real crash showed a calm yellow "Checking" pill with no error color and no admin-contact message — worse than the pre-PR behavior, which at least showed the raw text. Fixed by classifying `status.startsWith('Process exited')` as Error; added a regression test, verified red→green.
3. **`tests/e2e/export-page.e2e.ts` had 3 stale assertions** on the old "N exported, M skipped" text format, unrelated to any file this PR's Impact section listed. Updated to the new format.
4. **The new `machine-config-scanner-mode-persistence.e2e.ts` hardcoded the Windows/Linux keyboard shortcut**, which would fail on the `macos-latest` E2E CI job (Layout.tsx requires `metaKey` on Mac, not `ctrlKey`). Fixed to branch on `process.platform`.

## Important Issues (fixed)

- Restored a manual "Detect Cameras" retry button + loading/error state (lost during the #338 relocation — the original `CameraSettingsForm` had one).
- Generic save toast no longer stacks on top of a still-visible, undismissed restart-required notice.
- Export's success message now reads "(already present)" instead of the self-contradictory "(0 files)" when re-exporting files that already exist at the destination.
- Added tests for `restart()`/`checkHardware()` IPC rejection paths and "Test Connection" during manual camera entry — real, previously-uncovered code paths.
- Extracted the duplicated `HardwareStatus` type into `types/electron.d.ts`.
- Documented the restart-notice's page-local (not app-wide, lost on navigation) limitation in `design.md` as an accepted trade-off, with a same-commit text mitigation ("restart now, this notice won't reappear if you navigate away") rather than a much larger cross-component persistence rework.

## Not fixed (accepted as documented limitations, not regressions)

- Narrow race window between mount and camera-detection resolving where a fast typist could have a partial manual entry silently discarded — low probability, matches a pattern already present in the removed `CameraSettingsForm`.
- `MachineConfiguration.tsx`'s continued growth (~1000 lines, ~20 pieces of state) — flagged as a future decomposition candidate, out of scope for a "quick fixes" PR.
- `status.includes('ready')` substring check's long-term fragility — inherited unchanged from the pre-existing component, not introduced here.
- `originalConfig === null` edge case (only reachable if the initial `config.get()` rejects) silently skips the restart-notice check — narrow, and the fallback (generic toast only) is the safe default when the prior state is genuinely unknown.

---

eberrigan (commented):
## Round 2: Convergence Verification

Ran an adversarial verification pass against the round-1 fixes (b057b60), checking each against the actual current code rather than trusting the round-1 summary.

**7 of 8 fixes confirmed fully correct**: Basler help text restoration, crash classification, both stale-E2E fixes (export-page.e2e.ts + macOS keyboard shortcut), the Detect Cameras retry button, the double-banner-stacking fix, and the HardwareStatus type dedup — each verified against the actual rendered/tested behavior, not just presence of a diff.

**1 partial finding, now fixed (1fae184)**: the round-1 "(already present)" wording fix for Export's success message left one case unhandled — a `success` banner with `exportedScans === 0` (all selected scans no longer existed at export time, e.g. deleted elsewhere between page load and export) rendered as the misleading "0 scans exported (already present)", a false success claim with no indication anything was wrong. Now shows an explicit "No scans were exported — the selected scans may no longer exist" message instead, with a test covering it.

No new regressions found from the round-1 fixes themselves — `tsc --noEmit` and `npm run lint` clean, all directly-affected unit suites passing (75+ tests).

Review converged. No further findings outstanding beyond the already-documented, deliberately-accepted trade-offs (page-local restart notice, `MachineConfiguration.tsx` size, the narrow camera-detection race window).

eberrigan:
## Correction: admin shortcut leaked to Home's error message

During round-1 review fixes I added a hint to Home's Python-error message: "Admins: press Ctrl+Shift+, (Cmd+Shift+, on Mac) for hardware diagnostics." @eberrigan caught that this defeats Machine Configuration's entire access model — it's a shortcut-only, undiscoverable-by-design page, and that hint puts the shortcut directly in front of the technicians it's meant to be hidden from.

Fixed in `4195a38`, with TDD discipline (test written first, confirmed red against the bad text, then green after reverting):

- `PythonStatus.tsx`: reverted to the plain "Contact your administrator."
- `PythonStatus.test.tsx`: added a permanent regression assertion that this component's output never matches `/shift/i` or `/ctrl|cmd/i`, so this can't silently regress.
- `docs/CONFIGURATION.md`: found and fixed the identical flaw via a different channel — the admin-facing note said "relay the shortcut to lab technicians," which leaks the same information by word of mouth. Reworded to say technicians should know *who to contact*, not the shortcut.
- `design.md`: added a Correction note to the decision record.

All checks re-verified green locally (openspec validate --strict, prettier, tsc --noEmit, eslint, and the full PythonStatus + Home unit suites).

eberrigan:
## Manual smoke-test finding: self-contradictory export message

Found while manually smoke-testing (@eberrigan): re-exporting a scan whose files already exist at the destination showed **"1 scan exported (already present), 73 files skipped (already exist)"** — pairing "exported" with "already present" in the same clause reads as contradictory.

Also confirmed unrelated to this PR: the export folder structure itself (`scansExport()` in `database-handlers.ts`) mirrors each `Scan.path` DB field verbatim and wasn't touched here — a scan landing in a flat `scans/` folder without a date prefix reflects that scan's own stored `path` value (likely an older/local dev-db fixture predating the date/plant/uuid convention in `CaptureScan.tsx`'s `buildScanPath()`), not an export-logic change.

Fixed the message in `120f451` (TDD: test updated first, confirmed red, then green):
- When no new files are copied this run, the banner now reads "N scan(s) already exported — all K file(s) already present" instead of "N scan(s) exported (already present)".
- The normal case (some new files copied) is unchanged.
- Updated the matching unit (`Export.test.tsx`) and e2e (`export-page.e2e.ts`) assertions.

All checks green locally (prettier, tsc --noEmit, eslint, and the full Export unit suite — 15/15).

eberrigan:
## Manual smoke-test finding: nonsensical summary on total export failure

Found while continuing manual smoke testing (@eberrigan): when every selected scan fails outright (0 exported, 0 skipped — e.g. a source folder is missing), the banner still showed **"0 scans already exported — all 0 files already present"** directly above the failed-scan list. Technically not false, but confusing noise next to the actual failure info.

Fixed in `4e149fa` (TDD: test added first reproducing the exact wording, confirmed red, then green): a total-failure partial export now shows only the failed-scan list, no summary sentence. The success-only "no scans were exported" message and the normal exported/skipped summary are both unchanged.

Also confirmed two things reported during the same smoke-testing pass are **not bugs in this PR**:
- The ENOENT on a specific dev-db scan (`PLANT-002` / `salinity-tolerance-2025`) is `prisma/seed.ts`'s `scan2` fixture, which is deliberately seeded with no image files (its own comment: "no Image records... no point copying files it'll never reference"). Export correctly reports it as a failed scan rather than crashing.
- Duplicate `plant_id`+`experiment_id`+`wave_number`+`plant_age_days` is already hard-blocked in `CaptureScan.tsx` (`canStartScan` requires `!duplicateScanWarning`) — not just an advisory warning. The only gaps are outside that UI path (no DB-level unique constraint, so seed scripts/future bulk-import bypass it), pre-existing and out of scope here.

All checks green locally (prettier, tsc --noEmit, eslint, full Export unit suite — 16/16). CI running on `4e149fa` now.

